PR #1, opened 2014-12-31
Update to Rust master

Detegr:
Send trait is an unsafe trait and it cannot be derived anymore, therefore `unsafe impl` is needed. I got a panic when running tests (not always, but like 30% of the time) because the test tried to send data to a closed channel. So I changed a send to send_opt just to get rid of those panics (not that they would matter much). Also silenced all warnings.


passcod:
That's great work, thank you for this!

PR #2, opened 2014-12-31
Recursively watch subdirectories

Detegr:
Contrary to the documentation, I found out that `watcher.watch` does not watch subdirectories of `src`. Therefore I added `watch_recursive` to the Watcher trait and implemented it.

I'm not sure if this is necessary but at least on my machine I wasn't able to get any events from subdirectories. With this PR applied everything worked as I would expect.

### DIFF
--- a/src/inotify/mod.rs
+++ b/src/inotify/mod.rs
@@ -7,6 +7,7 @@ use std::io::IoErrorKind;
 use std::sync::{Arc, RWLock};
 use std::thread::Thread;
 use super::{Error, Event, op, Op, Watcher};
+use std::io::fs::{PathExtensions, walk_dir};
 
 mod flags;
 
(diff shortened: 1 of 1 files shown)


passcod:
My mistake, The Inotify impl is only file-recursive, not folder-recursive (I.e. it will watch the folder and every file it contains, but not descend into folders). The polling impl is properly folder-recursive, though. This needs to be solved at an implementation level, as different impls have differents behaviours (e.g. in linux kernel, fanotify is folder-recursive).

I'm going to leave this PR open until the bug is resolved and/or I write up a separate issue to track it.

Detegr:
My bad not reading through the polling impl. Fanotify seems plausible but it does not support create, delete and move events.

passcod:
Yes, I don't plan on making fanotify a recommended impl. Just implement recursion for inotify backend, then implement fanotify backend for those who want to use it even with its limitations.

Detegr:
I updated the pull request with an implementation for INotifyWatcher based on my previous implementation. Would this be okay to merge?

passcod:
This looks perfect, thank you very much :)

passcod:
I'll have to wait a bit before cutting a release and publishing a crate, though, as the new rust version broke a dependency :( (so I can't test).

PR #3, opened 2015-01-14
update to latest rust

kstep:
* `RWLock` renamed to `RwLock`
* `Thread::spawn()` now returns Thread struct, `Thread::scoped()` is for ThreadGuard now


passcod:
That was obsoleted, closing.

PR #5, opened 2015-02-15
use the new std::{fs, path}

blaenk:
This also requires an updated inotify-rs which you can find [in this branch](https://github.com/blaenk/inotify-rs/tree/new-io) (see the [PR](https://github.com/hannobraun/inotify-rs/pull/14)). It basically uses the new `std::io` for errors. You can use it like this:

``` toml
[target.x86_64-unknown-linux-gnu.dependencies.inotify]
git = "git@github.com:blaenk/inotify-rs.git"
branch = "new-io"
```

So `Path` is to `str` as `PathBuf` is to `String`, as I'm sure you already know. So things that contained `Path`s, like `HashMaps`, now contain `PathBuf`s.

One consequence of updating the inotify-rs package to use the new `std::io` is that there is no longer an `EndOfFile` variant of `ErrorKind`. Instead, `read` returns a `0` on EOF. I'm not sure exactly what should be done in this case, in the context of inotify, so in my inotify-rs branch I just returned an empty slice. This may change if you guys decide there's something else that should be done instead. Read [the PR](https://github.com/hannobraun/inotify-rs/pull/14) for more information.

One problem right now is that we forgot to implement `Hash` for `Path`; it's only implemented for `PathBuf`. This means that if you want to perform operations on a hashmap that contains `PathBuf` keys, but you only have a `&Path`, you have to convert it to a `PathBuf` in order to index the hashmap. [I filed a PR](https://github.com/rust-lang/rust/pull/22351) for this for rust, and for now added FIXME comments so we can remember to change this when it lands.

One slightly tedious change is that `fs::walk_dir` now returns an iterator that itself yields `Result`s. If the result is `Ok`, it yields a `DirEntry` type which you call `path()` on to get the actual Path.

As with my inotify-rs PR, I updated this to use the new `std::io::Error` instead of `IoError`.

Now, instead of `lstat()`, we use the `metadata()` method which has a `modified()` method which we can use.

### DIFF
--- a/src/inotify/mod.rs
+++ b/src/inotify/mod.rs
@@ -3,20 +3,20 @@ extern crate libc;
 
 use self::inotify_sys::wrapper::{self, INotify, Watch};
 use std::collections::HashMap;
-use std::old_io::IoErrorKind;
-use std::old_io::fs::{PathExtensions, walk_dir};
+use std::fs::{PathExt, walk_dir};
 use std::sync::mpsc::Sender;
 use std::sync::{Arc, RwLock};
 use std::thread::Thread;
 use super::{Error, Event, op, Op, Watcher};
+use std::path::{Path, PathBuf};
 
 mod flags;
 
 pub struct INotifyWatcher {
   inotify: INotify,
   tx: Sender<Event>,
-  watches: HashMap<Path, (Watch, flags::Mask)>,
-  paths: Arc<RwLock<HashMap<Watch, Path>>>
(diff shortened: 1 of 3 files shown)


passcod:
Merging to get it building, but will await the inotify-rs PR to merge before releasing a crate.

hannobraun:
FYI, I've just released the new version of inotify-rs to crates.io.

passcod:
Great :) I'll do the release dance tonight

PR #18, opened 2015-06-25
inotify: fixed delete is op::REMOVE, modify is op::WRITE.

payload:
Was other way around.


passcod:
Good catch!

Can you add yourself to the `authors` in the Cargo.toml ?

PR #22, opened 2015-07-11
Single file support for inotify

### LINKED ISSUE #16: Support watching single files
Currently the FSEvents backend supports watching single files, the inotify backend doesn't.

Go's fsnotify, Ruby's Listen, Facebook's Watchman and others also do not support watching single files.

__Phase 0__ of this feature is research:
- What is the reason behind lack of support in others?
- What are libraries that do have support?
- What are challenges for an implementation?
- Is this doable in a backend-independent manner?

If that last point can be answered by _yes_, another question is whether this behaviour would be better done in a separate package.

Yuhta:
Fix #16

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ authors = [
   "Michael Maurizi <mmaurizi@azavea.com>",
   "Pierre Baillet <oct@zoy.org>",
   "ShuYu Wang <andelf@gmail.com>",
+  "Jimmy Lu <gongchuo.lu@gmail.com>",
 ]
 
 description = "Cross-platform filesystem notification library"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,7 @@ version = "^0.2.11"
 
 [target.x86_64-apple-darwin.dependencies.fsevent-sys]
 version = "^0.1"
+
+[dev-dependencies]
+tempfile = "^1.1.0"
+tempdir = "^0.3.4"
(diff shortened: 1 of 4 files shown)


passcod:
Great work! Is inotify the only one that needs this, or do the others (polling) also need patching up?

Add yourself to the contributor array in the Cargo.toml :)

@maurizi Not sure if your work on the Windows implementation supports single file watching, is it possible, harder, easy ? Just to let you know we're apparently going to support that now.

Yuhta:
Patched and added `PollWatcher`; I don't have the environment to test `FsEventWatcher` but some one else can easily add it.

maurizi:
@passcod Windows only really has a notion of watching directories, so this will be harder on Windows, since we'll have to add a layer above this that filters out events not for that directory.  Since it looks like `PollWatcher` will need that as well, it might be a good idea to implement it generically.

passcod:
I'm going to merge this to _fix_ #16. FS watchers in general seem to only support watching folders, with exceptions.

So we'll have to figure an internal API, perhaps with traits, so a backend could:
- implement a `FileWatcher` to watch individual files, so we'd have to recurse to watch everything, or
- implement a `DirWatcher` to watch individual folders, so we'd have to recurse to watch subfolders, or
- implement a `RecursiveDirWatcher` to watch everything, or
- any combination of the above

and then Notify itself would take care of wrangling each backend to provide the same public API.

PR #25, opened 2015-07-26
Replace Path with AsRef<Path> in Watcher functions

frangio:
This should be a non-breaking change and it removes the need to call `Path::new` when calling `watch` or `unwatch`. This means `watcher.watch("/tmp/foo")` is now valid.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ authors = [
   "Pierre Baillet <oct@zoy.org>",
   "ShuYu Wang <andelf@gmail.com>",
   "Jimmy Lu <gongchuo.lu@gmail.com>",
+  "Francisco Giordano <frangio.1@gmail.com>",
 ]
 
 description = "Cross-platform filesystem notification library"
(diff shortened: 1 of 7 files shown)


passcod:
I've wondered how to do that... thank you very much for this :)
Add yourself to the Cargo.toml and I'll get this merged.

frangio:
No problem :)

frangio:
Hmm, I just realized that by removing the `&` now the methods are taking ownership of the argument when they really shouldn't. Give me a moment to sort it out.

frangio:
That should be it.

passcod:
Awesome :) I'll merge after work

frangio:
@passcod, sorry to bring this up again, but I think I was wrong with the last change I made (commit `bb3c978`). The method signature before that was better and is consistent with the Rust standard library. See for example http://doc.rust-lang.org/std/fs/struct.File.html#method.open. As for my concern regarding moving/borrowing, it's not a problem at all since the argument passed can still be a reference.

So I think commit `bb3c978` should be reverted.

passcod:
Done in 3340d74.

Issue #26, opened 2015-08-29
Tests hang if no events are received


jakerr:
I was doing some debugging of the tests trying to figure out why a CREATE event was delivered even though a file should have existed before the watch. My debugging uncovered a deadlock in `watch` when a `sleep_ms` is performed before it. Something about the `CFRunLoop` code I imagine but I couldn't track it down.

This causes the deadlock:
``` diff
diff --git a/tests/notify.rs b/tests/notify.rs
index 7fb5bb9..73fb23f 100644
--- a/tests/notify.rs
+++ b/tests/notify.rs
@@ -25,6 +25,7 @@ fn validate_watch_single_file<F, W>(ctor: F) where
   let mut file = NamedTempFile::new().unwrap();
   let (tx, rx) = channel();
   let mut w = ctor(tx).unwrap();
+  thread::sleep_ms(1000);
   w.watch(file.path()).unwrap();
   thread::sleep_ms(1000);
   file.write_all(b"foo").unwrap();
```

jakerr:
More of a hang than a deadlock really, putting the sleep here causes  the fsevent watcher to miss the modify events (maybe a bug, maybe just really hard to get the timing right on multi-threaded tests). Then `validate_recv` waits on an event that never occurs. The test should probably use rx.try_recv() in a loop with a timeout so if a failure results from a missing event it doesn't hang the test.

passcod:
Okay, the fix for this was apparently merged all the way back in #27, closing.

Issue #23, opened 2015-07-25
OSX - Watching 3 or more directories fails


compressed:
Hi,
I'm running OSX 10.10.4 and was trying to use the cargo watch tool, however it wasn't working. It seemed whenever the "benches" watch was added, cargo watch wouldn't watch anything...

Digging further, it seems something goes awry when adding the third watch. For the prior two watches, this condition is met (`fsevent.rs` line 62):

```rust
    if !self.is_running() {
      return;
    }
```
On the third try however, that condition isn't met, and it makes its way all the way to this match (`fsevent.rs` line 78):

```rust
      match context_info.done.recv() {
        Ok(()) => (),
        Err(_) => panic!("the runloop may not be finished!"),
      }
```
And never makes it past there...

Let me know if you need more information.

Also, if I try running cargo test, it fails:

```
~/rsnotify git:master ❯❯❯ cargo test                                                                                                                                                                                            ⏎
     Running target/debug/notify-a00f19333567126a

running 1 test
test fsevent::test_fsevent_watcher_drop ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured

     Running target/debug/notify-b65501780e7463a9

running 7 tests
test new_null ... ok
test new_fsevent ... ok
test new_recommended ... ok
test new_poll ... ok
test watch_dir_recommended ... FAILED
test watch_single_file_poll ... ok
test watch_single_file_recommended ... FAILED

failures:

---- watch_dir_recommended stdout ----
	thread 'watch_dir_recommended' panicked at 'assertion failed: `(left == right)` (left: `"/private/var/folders/b5/yh22bw792yj4n0q68vckrbjw0000gn/T/dir.KTsLmVgBAqyo"`, right: `"/var/folders/b5/yh22bw792yj4n0q68vckrbjw0000gn/T/dir.KTsLmVgBAqyo/dir1.W4y90yVhB8oP/dir11.EPrh1Cq6JFaE/.9ZIU5fW5tHKo"`)', tests/notify.rs:16


---- watch_single_file_recommended stdout ----
	thread 'watch_single_file_recommended' panicked at 'assertion failed: `(left == right)` (left: `"/private/var/folders/b5/yh22bw792yj4n0q68vckrbjw0000gn/T/.D36zoPQgBQr0"`, right: `"/var/folders/b5/yh22bw792yj4n0q68vckrbjw0000gn/T/.D36zoPQgBQr0"`)', tests/notify.rs:16



failures:
[...]

Issue #21, opened 2015-07-06
Errors when watcher backend is given broken symlinks


sunng87:
I'm on Linux kernel 4.1.0, rsnotify 2.1.0 and Emacs 24.5.

My code is as simple as:

```rust
fn _watch (p: &Path) -> Result<(), Error>{
    let (tx, rx) = channel();
    let mut watcher: RecommendedWatcher = try!(Watcher::new(tx));
    println!("watcher init");
    try!(watcher.watch(p));
    println!("watching");
    let _ = rx.recv();
    Ok(())
}
```

Every time I modified files in Emacs, `watcher.watch(p)` returns error:

```
Io(Error { repr: Os { code: 2, message: "No such file or directory" } })
```

If I `ls -al` the watched directory, I'm getting:

```
 $ ls -al examples/templates  
drwxr-xr-x 2 nsun nsun 4096 Jul  6 21:04 .
drwxr-xr-x 3 nsun nsun 4096 Jul  4 10:37 ..
lrwxrwxrwx 1 nsun nsun   31 Jul  6 21:04 .#index.hbs -> nsun@thinkless.17462:1435199339
-rw-r--r-- 1 nsun nsun  279 Jul  6 20:57 index.hbs
```

The `.#index.hbs` symbol link, as [the SO question](http://stackoverflow.com/questions/5738170/why-does-emacs-create-temporary-symbolic-links-for-modified-files) said, is Emacs lockfile. 

I'm not quite sure if this is an rsnotify issue, anything suggestion or workaround are welcomed.

JIghtuse:
It also happens if `watcher.watch` encounters broken symlink. Try to create one and you will see the same result. rsnotify just takes what walker gives it and throws it into `add_watch()`. What should it do instead?

passcod:
It should probably check that the path it's given is a directory or a regular file, and if it's a symlink it should attempt to dereference it, and only feed it to the watcher if either of those is the case. I imagine rsnotify also fails if invoked on, say, "special" files in `/dev`, `/sys`, or `/proc`.

passcod:
Fix shipped in v2.5.2.

PR #27, opened 2015-08-31
Tests on OSX.

octplane:
Tests now pass on OSX (necessary to reproduce #23 ).

- add OS conditionals for full path resolving (FSEvent API does send full resolved path to its callback), code courtesy of @jakerr.
- added some timeout during complex tests: FSEvent API sends *past events to its listener*, no matter what. The only solution is to wait.
- use @jakerr `validate_recv` method to match results in test cases and timeout (solves #26 ).
- enhance `validate_recv` to support aggregated events: `op::CREATE|op::REMOVE` will match `[op::CREATE, op::REMOVE]` (useful when your observing api can split or merge nearby events). Fixes a huge issue for OSX testing, has probably no impact on Linux.
- `pollwatch` test is now broken on OSX because of this full path stuff (worth another bug ? sibling of #21 ?)
- **untested** on linux. If the timeouts are too annoying, I still can refactor these to make runs faster on linux.

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -208,8 +208,8 @@ impl Watcher for FsEventWatcher {
       fsevent = FsEventWatcher {
         paths: cf::CFArrayCreateMutable(cf::kCFAllocatorDefault, 0, &cf::kCFTypeArrayCallBacks),
         since_when: fs::kFSEventStreamEventIdSinceNow,
-        latency: 0.1,
-        flags: fs::kFSEventStreamCreateFlagFileEvents,
+        latency: 0.0,
+        flags: fs::kFSEventStreamCreateFlagFileEvents | fs::kFSEventStreamCreateFlagNoDefer,
         sender: tx,
         runloop: Arc::new(RwLock::new(None)),
         context: None,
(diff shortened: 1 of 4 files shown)


passcod:
Back from holiday! This looks awesome :) I'll run/test on Linux and merge.

src/fsevent.rs
```diff
@@ -208,8 +208,8 @@ impl Watcher for FsEventWatcher {
       fsevent = FsEventWatcher {
         paths: cf::CFArrayCreateMutable(cf::kCFAllocatorDefault, 0, &cf::kCFTypeArrayCallBacks),
         since_when: fs::kFSEventStreamEventIdSinceNow,
-        latency: 0.1,
-        flags: fs::kFSEventStreamCreateFlagFileEvents,
+        latency: 0.0,
+        flags: fs::kFSEventStreamCreateFlagFileEvents | fs::kFSEventStreamCreateFlagNoDefer,
```
passcod: Out of curiosity, can you explain what that change does?

passcod:
We also need to have @jakerr added to the Cargo.toml.

passcod:
And finally, could you rebase on top of master? I've added OS X Travis CI testing, let's see what that does.

src/fsevent.rs
octplane: Defer will wait several event to aggregate them and deliver to the callback. This can mess tests (at least) and this seems to be a reasonable default, until somebody tells us this messes things. Same goes with `latency`.

PR #36, opened 2015-10-27
Fix event type in validate_watch_single_file text

aag:
Since the file is created before the watcher, and we only write to the existing file while it's being watched, we expect a WRITE event, not a CREATE event.

This fix changes the `watch_single_file_recommended()` test from always failing to intermittently failing
for me on Linux.


aag:
The test isn't behaving how I expect with FsEvent, so I'm going to close this PR for now. The problem could be related to #26.

PR #39, opened 2015-11-18
Add windows ReadDirectoryChanges support

### LINKED ISSUE #4: windows implementation
I know you already have this in your todo list, but I thought I would bring it to your attention that someone has provided the signatures for what I believe to be the necessary types and functions [here](https://github.com/retep998/winapi-rs/commit/df8b71132c31e93a82c6e06b45fc6ff53041d03e), in case you ever get around to this.

jmquigs:
Original code by @maurizi , with some fixes by me.

Fix #4


maurizi:
Great to see someone finish the work I started! :clap:

src/windows.rs
```diff
@@ -0,0 +1,306 @@
+extern crate libc;
```
retep998: This seems entirely unused.

src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #[macro_use] extern crate log;
 #[macro_use] extern crate bitflags;
 #[cfg(target_os="macos")] extern crate fsevent_sys;
+#[cfg(target_os="windows")] extern crate winapi;
```
retep998: Why is this both here and in `windows.rs`?

src/windows.rs
```diff
@@ -0,0 +1,306 @@
...
+extern crate kernel32;
+extern crate winapi;
+
+use winapi::{HANDLE, INVALID_HANDLE_VALUE, fileapi, winbase, winnt};
+use winapi::minwinbase::{OVERLAPPED, LPOVERLAPPED};
+use winapi::winerror::ERROR_OPERATION_ABORTED;
+use winapi::winnt::FILE_NOTIFY_INFORMATION;
```
retep998: All the definitions in the submodules of winapi are pub glob imported into the root module, so there's no need to refer to the submodules. It's only really there for code organization and documentation purposes.

src/windows.rs
```diff
@@ -0,0 +1,306 @@
...
+              | winnt::FILE_NOTIFY_CHANGE_SECURITY;
+
+    let monitor_subdir = if (&request.data.file).is_none() {
+      1
+    } else {
+      0
+    };
+
+    unsafe {
+        let mut overlapped: Box<OVERLAPPED> = Box::new(mem::zeroed());
```
retep998: The only time these get deallocated is when handling an event. When you shut down the thread you don't wait for the final set of callbacks to occur so you can cleanup the stuff, thus leaking any of these that were still in progress.

jmquigs: I can wait for the events on shutdown, but I suppose there is no way to check the handle to see if it is still valid in the completion routine without issuing another read call.  In that case, I guess I have to pass through some state flag to tell the routine that the server is shutdown and it shouldn't try any more reads.

Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = [
   "Antti Keränen <detegr@gmail.com>",
   "Gilbert Röhrbein <gilbert@ifsr.de>",
   "Jorge Israel Peña <jorge.israel.p@gmail.com>",
-  "Michael Maurizi <mmaurizi@azavea.com>",
+  "Michael Maurizi <michael.maurizi@gmail.com>",
```
passcod: Add yourself in there, too :)

appveyor.yml
```diff
@@ -0,0 +1,19 @@
...
+  - rust-nightly-%TARGET%.exe /VERYSILENT /NORESTART /DIR="C:\Program Files (x86)\Rust"
+  - SET PATH=%PATH%;C:\Program Files (x86)\Rust\bin
+  - SET PATH=%PATH%;C:\MinGW\bin
+  - rustc -V
+  - cargo -V
+
+build: false
+
+test_script:
+  - cargo test --verbose
```
passcod: Can we add a badge to the readme for this?

src/lib.rs
```diff
@@ -57,7 +60,8 @@ pub trait Watcher: Sized {
 
 #[cfg(target_os = "linux")] pub type RecommendedWatcher = INotifyWatcher;
 #[cfg(target_os = "macos")] pub type RecommendedWatcher = FsEventWatcher;
-#[cfg(not(any(target_os = "linux", target_os = "macos")))] pub type RecommendedWatcher = PollWatcher;
+#[cfg(target_os = "windows")] pub type RecommendedWatcher = ReadDirectoryChangesWatcher;
+#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))] pub type RecommendedWatcher = PollWatcher;
```
passcod: This is becoming a bit cumbersome... if anyone has an idea to make it so that we don't have to add to the attribute each time, please have a go :)

src/windows.rs
jmquigs: Nevermind, I get ERROR_OPERATION_ABORTED on the final read.  I can key off that...

appveyor.yml
jmquigs: I can add the badge, but I think it will be broken until you set up the project there.

passcod: I've just added it, so it should work :)

src/lib.rs
jmquigs: I removed the one from windows.rs

appveyor.yml
jmquigs: I see the CI build, but there is a project ID that I can't see that is needed to add the badge.  So I guess you'll have to add it :(  Also if you like, you can remove #4 from the todo list on the readme.
http://www.appveyor.com/docs/status-badges

passcod: Will do :)

passcod:
@maurizi @retep998 waiting on your approval/testing before merge :)

jmquigs:
@passcod, I'm going to investigate an alternate fix for the shutdown memory leak that @retep998 found.  The sleep I have there works for the common case, but if the completion routine isn't fired before sleep exits, it still leaks memory.  So if you created lots of watchers on a busy system and then shut them down, you'd probably leak some memory.

It will probably be a couple days before I can get that fix in.

jmquigs:
I made some good progress on that leak here: https://github.com/jmquigs/rsnotify/commit/02bca4456fc4005e2d56f75d3ab71adf8f2083ea.  I'm using semaphores to make sure that I wait for the final completion callback on each watcher.  Its not part of this PR; might be better to have a separate PR since there are so many changes here already.

src/windows.rs
```diff
@@ -0,0 +1,308 @@
...
+            Some(handle_event));
+
+        mem::forget(overlapped);
+    }
+}
+
+unsafe extern "system" fn handle_event(error_code: u32, _bytes_written: u32, overlapped: LPOVERLAPPED) {
+    // TODO: Use Box::from_raw when it is no longer unstable
+    let overlapped: Box<OVERLAPPED> = mem::transmute(overlapped);
+    let request: Box<ReadDirectoryRequest> = mem::transmute(overlapped.hEvent);
```
maurizi: I'm not sure what the minimum Rust version is for this project, but `Box::from_raw` has been stabilized as of 1.4

passcod: Latest stable.

jmquigs: I changed it to Box::from_raw

src/windows.rs
```diff
@@ -0,0 +1,307 @@
...
+      0
+    };
+
+    unsafe {
+        let mut overlapped: Box<OVERLAPPED> = Box::new(mem::zeroed());
+        // When using callback based async requests, we are allowed to use the hEvent member
+        // for our own purposes
+
+        let req_buf = request.buffer.as_mut_ptr() as *mut c_void;
+        let request_p: *mut c_void = mem::transmute(request);
```
retep998: This should be `Box::into_raw`

src/windows.rs
```diff
@@ -0,0 +1,307 @@
...
+            handle,
+            req_buf,
+            BUF_SIZE,
+            monitor_subdir,
+            flags,
+            &mut 0u32 as *mut u32,  // This parameter is not used for async requests
+            &mut *overlapped as *mut OVERLAPPED,
+            Some(handle_event));
+
+        mem::forget(overlapped);
```
retep998: Probably should pass `Box::into_raw(overlapped)` directly to `ReadDirectoryChangesW` instead of doing this.

tests/notify.rs
```diff
@@ -103,6 +148,11 @@ fn validate_watch_dir<F, W>(ctor: F) where
   let mut w = ctor(tx).unwrap();
 
   w.watch(dir.path()).unwrap();
+  // Windows needs some time for thread spinup before we start creating files.
```
retep998: Wouldn't it be better to have the watch function block until after `ReadDirectoryChangesW` is called?

src/windows.rs
jmquigs: Could do that; but in my other branch I check the return value here, because It can fail.  In the fail case you have to re-box the memory to get a drop:
https://github.com/jmquigs/rsnotify/blob/7dc1f555ff065154eb9e57134631e52bb958c6dc/src/windows.rs#L231
...maybe I should just include those changes here

tests/notify.rs
jmquigs: The watch server is in a separate thread, so it gets the message async via a channel.  Were you thinking of something like a busy wait here for a confirmation from the server?  In my other branch I've added a "MetaEvent" capability to the windows watcher.  It was intended for tests but could be used to implement this wait.  https://github.com/jmquigs/rsnotify/blob/7dc1f555ff065154eb9e57134631e52bb958c6dc/src/windows.rs#L41

jmquigs: watch() now waits for confirmation from the server, so the sleep is no longer necessary.

jmquigs:
I don't have any more pending changes for this PR.  It would be nice to add a dedup to the windows watcher are some point (its kinda spammy with writes to same file, for instance), but that should wait for another PR.

passcod:
@retep998 @maurizi Can either of you give the go-ahead for merging this?

jmquigs:
@passcod, its fine if you aren't comfortable merging this.  I'll keep the branch around in my fork anyway, since I need it for my project.

passcod:
Oh, I am totally comfortable merging this, but I don't have either the expertise to assess whether it's good code nor the setup to test it (appveyor takes care of some of that, but still). I'll probably end up merging it if nobody chimes in to tell me not to in a little bit, but I still want to give time for @retep998 and/or @maurizi to give it a look over and give me a LGTM.

Issue #37, opened 2015-10-28
inotify watcher thread never shuts down


benschulz:
The inotify watcher loop keeps going no matter what. Apparently the blocking call to `wait_for_events` does not get interrupted when the watcher is dropped/`close` is called. Given that, I don't know how to fix this. Should this be reported upstream?

passcod:
Yes.

PR #40, opened 2015-11-29
Replaced hand-rolled with mio-based event loop.

benschulz:
This is a follow-up to #37 and hannobraun/inotify-rs#28. Currently the inotify file descriptor is closed on the thread which drops the `Watcher`, for which the behavior is undefined. With these changes, a message is dispatched to the event loop, causing it to shut itself down.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,9 +39,11 @@ walker = "^1.0.0"
 
 [target.x86_64-unknown-linux-gnu.dependencies]
 inotify = "^0.1"
+mio = "^0.5"
 
 [target.x86_64-unknown-linux-musl.dependencies]
 inotify = "^0.1"
+mio = "^0.5"
 
 [target.x86_64-apple-darwin.dependencies]
 fsevent = "^0.2.11"
--- a/src/inotify/mod.rs
+++ b/src/inotify/mod.rs
@@ -2,63 +2,121 @@ extern crate inotify as inotify_sys;
 extern crate libc;
 extern crate walker;
 
+use mio::{self, EventLoop};
 use self::inotify_sys::wrapper::{self, INotify, Watch};
 use self::walker::Walker;
 use std::collections::HashMap;
 use std::fs::metadata;
 use std::path::{Path, PathBuf};
-use std::sync::mpsc::Sender;
+use std::sync::mpsc::{self, Sender};
 use std::sync::{Arc, RwLock};
-use std::thread;
+use std::thread::Builder as ThreadBuilder;
 use super::{Error, Event, op, Op, Watcher};
 
 mod flags;
 
-pub struct INotifyWatcher {
+const INOTIFY: mio::Token = mio::Token(0);
+
+pub struct INotifyWatcher(mio::Sender<EventLoopMsg>);
+
+struct INotifyHandler {
   inotify: INotify,
   tx: Sender<Event>,
   watches: HashMap<PathBuf, (Watch, flags::Mask)>,
   paths: Arc<RwLock<HashMap<Watch, PathBuf>>>
 }
 
-impl INotifyWatcher {
-  fn run(&mut self) {
-    let mut ino = self.inotify.clone();
-    let tx = self.tx.clone();
-    let paths = self.paths.clone();
-    thread::spawn(move || {
-      loop {
-        match ino.wait_for_events() {
-          Ok(es) => {
-            for e in es.iter() {
-              handle_event(e.clone(), &tx, &paths)
+enum EventLoopMsg {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -40,6 +40,11 @@ walker = "^1.0.0"
 [target.x86_64-unknown-linux-gnu.dependencies.inotify]
 version = "^0.1"
 
+[target.x86_64-unknown-linux-gnu.dependencies.mio]
+git = "https://github.com/carllerche/mio.git"
```
passcod: It is my understanding that libraries which depend on libraries _not in Crates.io_ cannot be published to Crates.io. Has this changed? If not, I cannot accept this PR until it points to a Crates.io-released version of `mio`.

src/inotify/mod.rs
```diff
@@ -116,73 +192,60 @@ fn handle_event(event: wrapper::Event, tx: &Sender<Event>, paths: &Arc<RwLock<Ha
...
+    INotify::init()
+      .and_then(|inotify| EventLoop::new().map(|l| (inotify, l)))
+      .and_then(|(inotify, mut event_loop)| {
+        let inotify_fd = inotify.fd;
+        let evented_inotify = mio::unix::EventedFd(&inotify_fd);
+
+        let handler = INotifyHandler {
+          inotify: inotify,
+          tx: tx,
+          watches: HashMap::new(), // TODO: use bimap?
```
passcod: Can you remove this TODO? I realise it is in the origin code, but it hasn't been relevant is a while, so might as well clean it up...

src/inotify/mod.rs
```diff
@@ -116,73 +192,60 @@ fn handle_event(event: wrapper::Event, tx: &Sender<Event>, paths: &Arc<RwLock<Ha
...
+                            INOTIFY,
+                            mio::EventSet::readable(),
+                            mio::PollOpt::level())
+          .map(|_| (event_loop, handler))
+      })
+      .map(|(mut event_loop, mut handler)| {
+        let channel = event_loop.channel();
+
+        ThreadBuilder::new()
+            .name("INotify Watcher".to_owned())
```
passcod: This looks like 4-space indentation while the rest of the code uses 2-space.

Cargo.toml
benschulz: I've never published a crate, but that makes a lot of sense. Do you mind if I keep this PR open in the meantime?

passcod: Sure. If there's an issue tracking this `mio` feature, could you link it here so we can keep track of when it makes it in a release?

passcod: The next `mio` milestone has a ETA for 3rd December, but I don't know whether `EventedFd` will be in there. I've asked on IRC for details.

benschulz: Thanks for looking that up/inquiring on IRC. Presumably 0.5.x will be branched off from master, so `EventedFd` should be included. I'll keep an eye on it either way.

passcod: IRC confirmed it :)

benschulz:
I've switched mio to version `^0.5`.

The windows builds are failing ("Could not compile `advapi32-sys`."). Should not have anything to do with this PR.

jmquigs:
I'll fix windows

Cargo.toml
```diff
@@ -43,6 +43,9 @@ version = "^0.1"
 [target.x86_64-unknown-linux-musl.dependencies.inotify]
 version = "^0.1"
 
+[target.x86_64-unknown-linux-gnu.dependencies.mio]
+version = "^0.5"
```
passcod: Also do this for `target.x86_64-unknown-linux-musl`

passcod:
You might have to rebase this.

benschulz:
Added mio for linux-musl, squashed and rebased.

Cargo.toml
```diff
@@ -43,6 +43,12 @@ inotify = "^0.1"
 [target.x86_64-unknown-linux-musl.dependencies]
 inotify = "^0.1"
 
+[target.x86_64-unknown-linux-gnu.dependencies.mio]
+version = "^0.5"
+
+[target.x86_64-unknown-linux-musl.dependencies.mio]
+version = "^0.5"
```
passcod: Could you use the style as shown by other stanzas? Add `mio = "^0.5"` to the relevant stanzas.

benschulz:
Done.

passcod:
Woohoo! Let's do this.

passcod:
Shipped in v2.5.2.

benschulz:
Cool, thanks. ;)

PR #42, opened 2015-12-03
Teach cargo configuration about musl libc.

cstorey:
It's possible to build binaries against a `--target` of `x86_64-unknown-linux-musl`, but unfortunately that doesn't match the current specifier for `x86_64-unknown-linux-gnu`. This patch just teaches cargo about the new configuration.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,9 @@ walker = "^1.0.0"
 [target.x86_64-unknown-linux-gnu.dependencies.inotify]
 version = "^0.1"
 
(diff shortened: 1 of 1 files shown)


passcod:
Great :) Out of curiosity, what are you using this for? Docker? Embedded? Some other usecase for a tiny libc?

cstorey:
Ah, it's just within docker for now. It seems neater to use a minimal operating environment like alpine linux than something like Debian when there's no need for the extra features.

passcod:
I'll release later today

PR #54, opened 2016-03-08
add fsevent{,-sys} dependencies to the i686-apple-darwin target

japaric:
r? @passcod


passcod:
[Released in 2.5.5](https://twitter.com/passcod/status/713901901918547968)

PR #70, opened 2016-06-03
Fixes #49: update inotify to v0.2

### LINKED ISSUE #49: Update to inotify 0.2
With 0.1, we build. With 0.2, we don't:

```
   Compiling notify v2.5.4 (file:///home/code/rust/notify)
src/inotify/mod.rs:76:17: 76:21 error: cannot move out of borrowed content [E0507]
src/inotify/mod.rs:76         let _ = self.inotify.close();
                                      ^~~~
error: aborting due to previous error
Could not compile `notify`.
```

julienw:
Uses an Option to store the inotify object.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,19 +38,19 @@ filetime = "^0.1.9"
 walkdir = "^0.1.5"
 
 [target.x86_64-unknown-linux-gnu.dependencies]
-inotify = "^0.1.12"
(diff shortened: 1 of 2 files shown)


julienw:
I tried it with a sample program and it works for me.

passcod:
Thank you very much for all these contributions!

julienw:
my pleasure :)

PR #71, opened 2016-06-03
Fixes #55: Implement StdError for notify::Error

### LINKED ISSUE #55: std::error::Error implementation for notify::Error
Consider implementing `std::error::Error` trait for `notify::Error` because it helps with interoperability with other code. Currently I'm unable to use the error with `{}` pattern when formatting a string, for example.

julienw:


julienw:
This is useful when we want to chain this error inside another error and return a cause :)

passcod:
Can you rebase this?

julienw:
Just rebased !

PR #72, opened 2016-06-03
Add a Result alias to allow for an easier use in lib users

julienw:


julienw:
This makes it easier to return a result with this error by returning `notify::Result<RecommendedWatcher>`. instead of `Result<RecommendedWatcher, notify::Error>`. Like `fmt::Result` or `io::Result`.

PR #74, opened 2016-06-04
Only register directories for watching

vandenoever:
Directories also report the events for the files in the directories.

### DIFF
--- a/src/inotify/mod.rs
+++ b/src/inotify/mod.rs
@@ -95,6 +95,18 @@ impl mio::Handler for INotifyHandler {
     }
 }
 
+/// return DirEntry when it is a directory
+fn filter_dir(e: walkdir::Result<walkdir::DirEntry>) -> Option<walkdir::DirEntry> {
+    if let Ok(e) = e {
+        if let Ok(metadata) = e.metadata() {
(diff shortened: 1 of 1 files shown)


passcod:
Walk me through this. What does your change do? Does it still support calling `add_watch` on a file?

vandenoever:
For normal files there is no need for a watch, so for files instead of calling add_watch, Ok(()) is returned.

vandenoever:
I see now that this patch does not yet support calling add_watch_recursively to register a watch on a single file.

passcod:
This function originally has two scenarios:

- If Path is a file, watch it
- If Path is a dir, watch dir, all subdirs and subfiles

We want:

- If Path is a file, watch it
- If Path is a dir, watch dir and all subdirs only

Your patch does, even the updated version:

- If Path is a file, fail silently
- If Path is a dir, watch dir, all subdirs and subfiles

So, see, what your patch does is silently break a valid use case, and fail to actually address the problem.

vandenoever:
I uploaded a new diff in pull request #76 . I'm on a very bad connection and have a hard time updating the diff on this PR.

PR #75, opened 2016-06-05
Simplify target dependencies in Cargo.toml

nicokoch:


passcod:
This actually works? Wooo! Thanks a bunch!

PR #81, opened 2016-06-22
Watch only new directories in the monitor example

HessiJames:
The current example can be confusing since directories report events for their files anyway.

If the newly created files are watched on top of their parent directory, some events get emitted multiple times and in some cases out of order. Since rsnotify condenses events to only 6 possible values, it can be hard to understand what's going on when monitoring a directory.


passcod:
Good catch! Also, I took the 6 events from other libraries (Go's fsnotify, iirc), but if you have a proposal to expand that sensibly, then please :)

PR #83, opened 2016-06-28
Implement Send and Sync for OSX's FsWatcher and Windows' ReadDirectoryChangesWatcher (closes #82)

### LINKED ISSUE #82: MacOSX' RecommendedWatcher (FsEventWatcher) doesn't implement Send
This is misleading because other watchers are Send and so can be moved between threads.

Here is an example of an error I get in MacOS X only:
```
src/lib.rs:643:9: 643:22 error: the trait bound `*mut libc::c_void: std::marker::Send` is not satisfied [E0277]
src/lib.rs:643         thread::spawn(move || {
                                                                                                          ^~~~~~~~~~~~~
src/lib.rs:643:9: 643:22 help: run `rustc --explain E0277` to see a detailed explanation
src/lib.rs:643:9: 643:22 note: `*mut libc::c_void` cannot be sent between threads safely
src/lib.rs:643:9: 643:22 note: required because it appears within the type `notify::FsEventWatcher`
src/lib.rs:643:9: 643:22 note: required because it appears within the type `FsWatcher`
```

julienw:


julienw:
Not tested on a real Mac yet.

tests/notify.rs
```diff
@@ -230,3 +230,31 @@ fn new_recommended() {
...
+
+// if this test builds, it means RecommendedWatcher is Sync.
+#[test]
+fn test_watcher_sync() {
+    use std::sync::{ Arc, RwLock };
+
+    let (tx, _) = channel();
+
+    let watcher: RecommendedWatcher = Watcher::new(tx).unwrap();
+    let watcher = Arc::new(RwLock::new(watcher));
```
julienw: I had an hard time coming up with an idea to test this. In the end I use RwLock so that I can get &mut references and be able to call `watch`. A Mutex wouldn't work as it doesn't mandate `Sync`.

passcod:
Well, seems to work on Travis's Mac runners, at least :+1:

julienw:
Do you know why the builds in AppVeyor are failing ?

julienw:
Ah I know: same issue on Windows, I guess. I can have a look as well.

julienw:
OK, I'm not sure I understand why we use this semaphore at all instead of using Rust's synchronization mechanisms (like a Mutex on ()), but after some analysis I think it's safe to implement Send/Sync for the Windows Watcher as well.

tests/notify.rs
jwilm: You could write a function like

```rust
fn assert_sync<T: Sync>() {}
```

and call it in the test

```rust
fn test_watcher_sync() {
    assert_sync::<RecommendedWatcher>();
}
```

It's not a test in the usual sense since it will actually fail at compile time, but that seems ok :).

maurizi:
> OK, I'm not sure I understand why we use this semaphore at all instead of using Rust's synchronization mechanisms (like a Mutex on ()), but after some analysis I think it's safe to implement Send/Sync for the Windows Watcher as well.

It is using a Windows-provided semaphore because the thread needs to wait on the semaphore for Windows to be able to wake the thread up when file events occur.  The semaphore gets passed between threads already (via transmuting back/forth from a u64) so it shouldn't be any less safe to make the Windows watcher Send/Sync if thats the only thing preventing it.

julienw:
> It is using a Windows-provided semaphore because the thread needs to wait on the semaphore for Windows to be able to wake the thread up when file events occur. 

But I don't understand how it's used. But likely just because I'm a complete newbie in Windows-based programmation.

> The semaphore gets passed between threads already (via transmuting back/forth from a u64)

Yeah that's what I've seen :)

tests/notify.rs
julienw: I thought it was better to have a test that tried to do something useful than something clearly dummy. Maybe in the future we could also augment it to actually see if the watcher works. (tell me if you'd like me to do this now already)

julienw:
OK everything is green. But I still haven't tried _for real_ on Mac and Windows, as I don't have these systems at all. Does one of you have some code and access to these systems ? Or are we confident the basic functionality works thanks to other tests ?

passcod:
I'm fine if the tests pass, but maybe @maurizi can have a go, if possible.

passcod:
@HessiJames Given your PR re osx, is it possible you'd have a "real mac" to test this with?

HessiJames:
No, just a VM, sorry.

passcod:
Ok I'll just merge this @julienw

julienw:
thx :) we can just backout in case somebody complains.

passcod:
Released in 2.6.2.

PR #85, opened 2016-07-03
Remove reference counter and lock from paths map

HessiJames:
paths is never shared with another thread, so accessing it is always safe.


passcod:
Well that simplifies things! Thank you

PR #86, opened 2016-07-03
Refactor and add more tests

HessiJames:
I have started to write some tests. They are not complete and not all tests succeed. I disabled ("ignored") the ones that don't work for now. And I would like to add some more tests, so this is just the first batch.

~~I set up VMs with windows and os x to test it and there are quite a few differences on various platforms. In order to deal with all the minor differences I had to use a lot of conditionally compiled asserts. To simplify things I use the stmt_expr_attributes feature, which requires rust nightly.~~

~~I also changed the Windows FILE_ACTION_RENAMED_NEW_NAME event from RENAME to CREATE event in order to match the Linux behavior. This is a breaking change, so I set the version number to 3.0.0. I didn't find a changelog file though.~~

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -116,7 +116,7 @@ impl ReadDirectoryChangesServer {
 
             unsafe {
                 // wait with alertable flag so that the completion routine fires
-                let waitres = kernel32::WaitForSingleObjectEx(self.wakeup_sem, 500, TRUE);
+                let waitres = kernel32::WaitForSingleObjectEx(self.wakeup_sem, 100, TRUE);
                 if waitres == WAIT_OBJECT_0 {
                     let _ = self.meta_tx.send(MetaEvent::WatcherAwakened);
                 }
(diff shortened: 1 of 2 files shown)


passcod:
Is it possible to split this pr in two so the breaking change is isolated?

.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: rust
+rust:
+  - nightly
```
passcod: Does this mean this PR makes Notify only build on Nightly? We aim to track Stable, so it would have to wait.

passcod: I'd really like to track stable, and that includes tests. What would be needed to run tests on stable? Where is the `stmt_expr_attributes` feature at, in terms of getting to stable? Can we avoid it?

HessiJames: Yes, it can be avoided, as it turns out using the cfg! macro works even better.

passcod:
Is this complete or do you want to add some more to it? I.e. can you say when this is ready to merge?

HessiJames:
I think it's ready to be merged. As I said, I'm working on some more changes, but I don't think it's a good idea to put them all into one merge request.

Issue #68, opened 2016-05-02
Allow backends to provide their own recursive watcher


passcod:
For example, if a backend has a better way to watch directories recursively, we should let it do that.

Issue #61, opened 2016-04-30
does inotify leak when monitored objects are deleted?


blaenk:
I'm wondering what happens if a monitored object is deleted. I'm wondering if it would remain/leak/linger in the `watches` map. Perhaps we can watch for the `IN_DELETE_SELF` event for each watched path, and if triggered, we remove the entries? Not sure if that'd work.

```
           IN_DELETE_SELF
                  Watched file/directory was itself deleted.  (This event
                  also occurs if an object is moved to another filesystem,
                  since mv(1) in effect copies the file to the other
                  filesystem and then deletes it from the original
                  filesystem.)  In addition, an IN_IGNORED event will
                  subsequently be generated for the watch descriptor.
```

Also from TLPI:

> An `IN_IGNORED` event is generated when a watch is removed. This can occur for two reasons: the application used an `inotify_rm_watch()` call to explicitly remove the watch, or the watch was implicitly removed by the kernel because the monitored object was deleted or the file system where it resides was unmounted.

So I think monitoring for `IN_IGNORED` should be enough. When we receive it, we remove the entries from our own data structures.

jwilm:
That sounds right

passcod:
When this happens, should we also notify the user? How?

jwilm:
@passcod you mean in addition to the REMOVE event? We could deliver and IGNORE event as well.

blaenk:
The way we'd discover this in rsnotify ourselves is itself a notification. That is, `IN_DELETE_SELF` or `IN_IGNORE`. I don't think we currently propagate `IN_IGNORE` because it probably doesn't have a "generic" equivalent on the other back-ends. My understanding from what I quoted is that first we would get an `IN_DELETE_SELF` which we propagate to the user as usual, then when we receive an `IN_IGNORE` we should remove the entries from the structures. I think it's reasonable for them to conclude that if it was deleted, they won't be getting any events for it anymore. Or if we're okay with introducing an IGNORE event then we can just propagate that as well.

passcod:
Right, no, just sticking to the current behaviour should be fine :)

Issue #60, opened 2016-04-30
INotifyWatcher recursively adds watches but does not recursively remove


jwilm:
Here's where the INotifyHandler gets requests to add/remove watches.
```rust
            EventLoopMsg::AddWatch(path, tx) => {
                let _ = tx.send(self.add_watch_recursively(path));
            }
            EventLoopMsg::RemoveWatch(path, tx) => {
                let _ = tx.send(self.remove_watch(path));
            }
```

Watches are definitely done recursively (that function does a `WalkDir` and `add_watch` for each entry). Remove watch only attempts to remove the given path.

```rust
    fn remove_watch(&mut self, path: PathBuf) -> Result<(), Error> {
        match self.watches.remove(&path) {
            None => Err(Error::WatchNotFound),
            Some(p) => {
                let w = p.0;
                match self.inotify.rm_watch(w) {
                    Err(e) => Err(Error::Io(e)),
                    Ok(_) => {
                        // Nothing depends on the value being gone
                        // from here now that inotify isn't watching.
                        (*self.paths).write().unwrap().remove(&w);
                        Ok(())
                    }
                }
            }
        }
    }
```

Just for completeness, here's the `add_watch_recursively` implementation.

```rust
    fn add_watch_recursively(&mut self, path: PathBuf) -> Result<(), Error> {
        match metadata(&path) {
            Err(e) => return Err(Error::Io(e)),
            Ok(m) => {
                if !m.is_dir() {
                    return self.add_watch(path);
                }
            }
        }

        for entry in WalkDir::new(path)
                         .follow_links(true)
                         .into_iter()
                         .filter_map(|e| e.ok()) {
            try!(self.add_watch(entry.path().to_path_buf()));
        }

        Ok(())
    }
```

blaenk:
Nice catch. I haven't had a chance to work on rsnotify in a while, but one thing that may be important to keep in mind is that, to remove the watches, we probably shouldn't just recursively traverse the "root" since perhaps new directories were removed since then.

For example, we perhaps could have recursively traversed the root and for each directory found, check if it's in `watches`, if so, `remove_watch(that_path)`. However, one problem is that as far as I know, we make no distinction as to whether a watch is recursive or not. So given some path that we want to remove, we won't know whether we should recursively remove it or not.

[...]

Issue #20, opened 2015-06-26
PollWatcher improvements


passcod:
From the PollWatcher code:

```
// TODO: more efficient implementation where the dir tree is cached?
// TODO: populate mtimes before loop, and then handle creation events
// TODO: handle deletion events
// TODO: handle chmod events
// TODO: handle renames
// TODO: DRY it up
```

passcod:
The first two (the work done in #88), have been released in 3.0.0.

passcod:
This will be all done with the new poll-tree backend in v5.

dhiemaz:
@passcod, this issue solved yet?

passcod:
Nope: https://github.com/notify-rs/notify/blob/main/src/poll.rs#L59-L60

PR #88, opened 2016-07-09
Normalize the recursive watch behavior of the various back-ends

HessiJames:
This patch normalizes the recursive behavior of the various back-ends when adding watches.
Currently inotify initially watches directories recursively, but doesn't watch new directories and cannot unwatch directories recursively.
Windows and FSEvent watch directories recursively and automatically watch new directories.

With this patch, all back-ends can watch directories recursively (while automatically watching new directories) or non-recursively depending on a new flag for the watch(..) function.
It also fixes a memory leak in the inotify back-end.

I implemented CREATE and DELETE for PollWatcher in order to make it compatible with the new tests.
And I increased the PollWatcher's default delay to 10 seconds. I guess that the default value of 10 ms was a mistake, since that seems a bit low. Plus the PollWatcher has only a time resolution of 1 second anyway.

I also moved some code from tests/notify.rs. to the new files tests/watcher.rs and tests/utils.rs in order to reduce its size.

Here are some issues, this patch might affect:
Fixes: #60
Fixes: #61
Implements some features from: #20
Fixes: #68

### DIFF
--- a/examples/monitor.rs
+++ b/examples/monitor.rs
@@ -1,20 +1,8 @@
 extern crate notify;
 
-use notify::op::CREATE;
-use notify::{RecommendedWatcher, Watcher};
+use notify::{RecommendedWatcher, Watcher, RecursiveMode};
 use std::path::Path;
 use std::sync::mpsc::channel;
-
-#[cfg(target_os="linux")]
-fn add_watch<P: AsRef<Path>>(watcher: &mut RecommendedWatcher, path: P) -> notify::Result<()> {
-    watcher.watch(path)
-}
-
-// OS X and Windows already monitor the entire file tree, including newly created directories
-#[cfg(not(target_os="linux"))]
-fn add_watch<P: AsRef<Path>>(watcher: &mut RecommendedWatcher, path: P) -> notify::Result<()> {
-    Ok(())
-}
 
 fn watch<P: AsRef<Path>>(path: P) -> notify::Result<()> {
     // Create a channel to receive the events.
@@ -26,20 +14,12 @@ fn watch<P: AsRef<Path>>(path: P) -> notify::Result<()> {
 
     // Add a path to be watched. All files and directories at that path and
     // below will be monitored for changes.
-    try!(watcher.watch(path));
+    try!(watcher.watch(path, RecursiveMode::Recursive));
 
     // This is a simple loop, but you may want to use more complex logic here,
     // for example to handle I/O.
     loop {
         match rx.recv() {
-          Ok(notify::Event{path:Some(path), op:Ok(CREATE)}) => {
-              println!("{:?} {:?}", CREATE, path);
-              if path.is_dir() {
-                  if let Err(e) = add_watch(&mut watcher, path) {
-                      println!("error adding watch {}", e);
-                  }
-              }
-          },
           Ok(notify::Event{path:Some(path), op:Ok(op)}) => {
               println!("{:?} {:?}", op, path);
           },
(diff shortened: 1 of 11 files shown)


passcod:
Very very cool! Really awesome work

passcod:
This all looks good. Again, many thanks and great work on this.

passcod:
Do you have any more awesomeness you're working on before I merge all this and cut 3.0.0?

HessiJames:
Glad you like it. There still are some issues I would like to fix, so don't release 3.0.0 just yet.
I would suggest you merge this PR now, but what really matters to me is that master doesn't change, because I would like to avoid merge conflicts. You can expect new PRs soon.

passcod:
In that case, can you merge or rebase your other PR on top of this, so I
can merge both now?

passcod:
Actually, do you want commit rights on the repo for this? I can remove you afterwards if you don't want the long term, but that should make it easier for this code effort :)

HessiJames:
Well, that's up to you. I certainly wouldn't mind having commit rights. But you should make it clear what kind of changes I should push directly and which you want to review first.

My next plans are to better detect renames by adding a cookie (like in inotify) to the events. And then make sure that all events reported are the same for at least Linux and Windows. Doing that for OS X is somewhat difficult. And I want to clean up the tests some more.

passcod:
My idea was that given what you've been doing so far has been _excellent_, I tentatively trust you to commit good things. Several things, in no particular order:

- This is git. I have a copy of the entire history locally. Giving you commit rights is no big risk to me, as I can remove it at any time and revert anything back to how it was. So I'm not worrying there. (The worst thing you could possibly do is deface the repo in such a way that it brings negative publicity, I guess? But even then, again, easy to undo, and publicity is publicity! ;P)
- I can comment on commits, so I can discuss changes in _post_ rather than in pull requests, if needed… and in the meantime you can go forward. I'm not above history rewrites if big chunks need to be reverted, although I don't expect that to happen.
- Obviously you can't change the project metadata (name, license, etc), and you can't replace it all by a library that sings the 99 bottle song in albanian morse code. But apart from that… more usefully, #24 has a big picture view of what my goals have been with Notify from the start. The discussion attached to it is a bit dated, but the bullet points are all still current.
- I think at this point changing the entire public API around would be detrimental, unless there's a very good reason. So, if you think that's something that should happen, that's a thing to discuss beforehand.
- You have, with your PRs, **essentially refactored most of the library**. In fact, the first thing you can do with commit rights is add yourself to the author list in the Cargo.toml. I mean, I'm still project owner, and there's other people who have done a lot of work, and I'll still be maintaining and managing this library in a year, two years… even after you don't actually come around anymore, if that happens. But this is a young project and you've done a lot, albeit in a short time; you are therefore actually a part of this. For how long, who knows; and it matters not.
- Ah, something I hadn't thought of: please don't insert backdoors in the code, mmkay? Not that you would respect that if you were going to, but yeah, please don't. Easter eggs are fine; if that can be managed sensibly in such a library, that would be an achievement in itself.
- If you want commit rights to be revoked at any point: well, you can do it yourself anyway, but please do tell when you do it, just so I know.
- Obviously if you're not sure about something you want to change or want a second pair of eyes, ask.

Unrelated, may I ask your preferred pronouns/gender so I use the right ones when I talk about you or your work here? The anime avatar is rather opaque to guessing :)

And finally, and on topic: during week days nowadays I have access to a real OS X box, so if you need something tested do ask.

HessiJames:
Ok, I think I got the gist :) ... no backdoors then ...
Thanks for the invite and fyi I'm male.

PR #93, opened 2016-08-13
Add IN_CLOSE_WRITE Event

xieke91:
The IN_MODIFY event is emitted on a file content change (e.g. via the write() syscall) while IN_CLOSE_WRITE occurs on closing the changed file. It means each change operation causes one IN_MODIFY event (it may occur many times during manipulations with an open file) whereas IN_CLOSE_WRITE is emitted only once (on closing the file).

### DIFF
--- a/src/inotify/mod.rs
+++ b/src/inotify/mod.rs
@@ -161,6 +161,9 @@ impl mio::Handler for INotifyHandler {
                                     if event.is_modify() {
                                         o.insert(op::WRITE);
                                     }
+                                    if event.is_close_write() {
+                                        o.insert(op::CLOSE_WRITE);
+                                    }
                                     if event.is_attrib() {
                                         o.insert(op::CHMOD);
                                     }
@@ -253,7 +256,7 @@ impl INotifyHandler {
     }
 
     fn add_single_watch(&mut self, path: PathBuf, is_recursive: bool, watch_self: bool) -> Result<()> {
-        let mut flags = flags::IN_ATTRIB | flags::IN_CREATE | flags::IN_DELETE |
+        let mut flags = flags::IN_ATTRIB | flags::IN_CREATE | flags::IN_DELETE | flags::IN_CLOSE_WRITE | 
                         flags::IN_MODIFY | flags::IN_MOVED_FROM | flags::IN_MOVED_TO;
 
         if watch_self {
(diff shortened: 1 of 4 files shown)


HessiJames:
Thank you for your patch.
I think it makes sense to forward `IN_CLOSE_WRITE` since it is a very useful event.
The biggest problem with `IN_CLOSE_WRITE` is that it is only available with inotify. But given that it can be very useful, it might be worth adding anyway. Though it has to be made clear in the docs that it is only available on Linux.
Also the tests need to be updated (I can do that if you want).
And some more nitpicking: I would call it `CLOSE_WRITE` (with an underscore) and you should add a changelog entry.
What do you think, @passcod ?

xieke91:
ok!

HessiJames:
@xieke91 Do you know that you can update your pull requests, you don't have to close them and open new ones.
Just push new commits to your forked repository and they will show up in the pull request.

And if you want to have a nice looking git log, you can use [git rebase](https://help.github.com/articles/about-git-rebase/) to rewrite your local history and use `git push --force` to forcefully push your changes to your forked repository.

xieke91:
sorry,I am not very familiar with git and I'm already update it with CLOSE_WRITE.

xieke91:
I try to edit tests but something wrong in windows and macos,maybe you can do it.

HessiJames:
I don't think that the test that's failing in travis build nr. 215.1 has anything to do with your patch. I'll have a look at that.
I think it looks ok, but since I'm not the maintainer, I won't merge it, lets see what @passcod has to say.

passcod:
Yep, looks good. I think kqueue (BSD's inotify) may support an analog to IN_CLOSE_WRITE, so it may not be a linux-only thing in the long term.

@HessiJames I'll have a closer look (finally! it's been a busy few weeks) at debounce, but do you think CLOSE_WRITE could be integrated somehow? Like, can we have some way of using CLOSE_WRITE on linux and a debounced MODIFY on other platforms to watch for "a file has been modified"? Just thinking out loud, it's probably infeasible, hah.

HessiJames:
I thought about that as well, but in the end I think it would only complicate things. The debounce module debounces more than just write events (eg. "safe saves"), so using CLOSE_WRITE would introduce inconsistencies not only across platforms but also across events.
Actually I think the use-cases for CLOSE_WRITE in notify are quite narrow. But including it doesn't hurt much and it gives developers the possibility to implement such a "custom debounce" logic if they really need it.

passcod:
That makes sense 👍

PR #104, opened 2016-12-21
update dependency to inotify 0.3.0

astraw:
Hi, I updated the dependency on inotify from 0.2.3 to 0.3.0. All test pass on my machine.


dfaust:
Thank you.

Issue #105, opened 2016-12-29
`.watch()` panics on macOS when the target doesn't exist


passcod:
See passcod/cargo-watch#50.

PR #111, opened 2017-01-22
Fix  850ff40095e02771a0b7b0518ef9ba2678200096

Np2x:
With 850ff40095e02771a0b7b0518ef9ba2678200096 I introduced a bug,
when a path couldn't be found the watcher wasn't started again. This
should fix issue #105 correctly.


dfaust:
Thanks!

Issue #112, opened 2017-02-20
Rename Chmod event to Metadata


alexbowers:
When you `touch` a file, it currently files the chmod, which sounds incorrect (chmod is about permissions on *nix).

An alias to metadata would help code read a little better

dfaust:
I like the idea. But adding another variant to `DebouncedEvent` is not a good solution because we would have to send metadata change events twice and every API user would have to handle them (or get a compiler warning). So we would have to replace the `Chmod` variant.

alexbowers:
Is deprecating the `Chmod` variant an option perhaps, and changing its name to `Metadata` in a future release?

passcod:
Yes. The chmod event is always about file metadata, and never limited to just permissions, so that's a good idea. This is both applicable to Chmod (in DebouncedEvent) and CHMOD (in Op).

I think we should mark those as deprecated and mention that they'll be renamed. Then the actual change should be done whenever the next major goes through. That might be a while, though.

alexbowers:
@passcod Has this been implemented? #114 mentions it was, but I can't see it in code.

passcod:
Err. Hmm. No, I don't think I've done this. The spec in #117 is basically where I wanted to go, with the current release as a "final 4.x version". As in, I could deprecate this particular thing, but if the API is going to radically change anyway... 🤷

Unfortunately I've mostly lost interest in this project, and that plus being very busy at work and elsewhere has my attention _gone_. So notify is in a holding pattern of "it works."

passcod:
This is done in v5, which also brings a different approach to event classification.

PR #114, opened 2017-02-26
Replace time with std::time

serprex:
Cargo will still pull ``time`` in as a dependency at times through mio, but mio 0.6 doesn't depend on ``time``

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ categories = ["filesystem"]
 [dependencies]
 bitflags = "^0.7.0"
 libc = "^0.2.4"
-time = "^0.1.34"
 filetime = "^0.1.9"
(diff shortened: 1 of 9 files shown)


dfaust:
Thank you very much. I looked through the code and it looks good.

The biggest change I could find is that the poll watcher now uses the monotonic - but not steady - [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) instead of the high-resolution performance counter from `time`. This shouldn't change the behavior of the poll watcher, though, because the time is only used to compare `last_check` time-stamps. The polling interval is still set by `thread::sleep()`.

passcod:
I'll release this as a patch-level change later this week, along with the deprecation from #112.

Issue #131, opened 2017-11-20
crash on macOS (fsevent) after watching same file twice, then unwatching


cmyr:
The following snippet consistently crashes for me:

```rust
        let (tx, rx) = channel();
        let path = PathBuf::from("/usr/local/bin/git");
        let mut w = watcher(tx, Duration::from_secs(1)).unwrap();
        w.watch(&path, RecursiveMode::NonRecursive).unwrap();
        w.watch(&path, RecursiveMode::NonRecursive).unwrap();
        w.unwatch(&path).unwrap();
    }
```

~I'm not sure what I should expect if attempting to explicitly watch the same file twice; it might be reasonable to either return an error from `watch`?~

edit: I'm now wondering if this crash isn't due to modifying the underlying cfarray during iteration ([here](https://github.com/passcod/notify/blob/master/src/fsevent.rs#L123)) , and ending up out of bounds.

I can put together a PR for this, but I wonder if we shouldn't also have some clearer expectations around how to behave with duplicate entries? Since this is going to be (likely) very implementation dependent, I suspect we should disallow it? I'll defer to @passcod on that.

passcod:
My plan would be more to silently ignore duplicate paths. This is how it's shaping up to be in the next version. ("frontend" would never pass overlapping or duplicate paths to "backend")

cmyr:
@passcod sounds good. I expect to be making quite extensive use of this crate in [xi](https://github.com/google/xi-editor) over the next little while, and I should have some downtime in december; if you're doing a rewrite and want a hand with anything feel free to ping me.

PR #132, opened 2017-11-20
Fix fsevent crash (mutating array in iteration)

cmyr:
This is a quick fix for #131, although I think some questions remain around how we should be considering duplicate watch events, in general.


passcod:
Looks sensible to me.

The tests failing macOS in CI are existing failures, so merging.

cmyr:
@passcod great, thanks for the quick merge. Also just wanted to say your tests have been really useful to me today in trying to figure out how to write some tests of my own.  Thanks! 😄 ✨

Issue #118, opened 2017-05-16
Deadlock due to race in FsEventWatcher


michaelfairley:
Because the runloop thread [signals that the runloop has started _before_ `CFRunLoopStart` is called](https://github.com/passcod/notify/blob/80594cf6aec242a0986c45ffc31ff692d30d1698/src/fsevent.rs#L208-L210), it's possible for [`CFRunLoopStop` to get called](https://github.com/passcod/notify/blob/80594cf6aec242a0986c45ffc31ff692d30d1698/src/fsevent.rs#L102) _before_ `CFRunLoopStart`, causing the runloop to never stop and thus [the `done` channel to be waited on forever](https://github.com/passcod/notify/blob/80594cf6aec242a0986c45ffc31ff692d30d1698/src/fsevent.rs#L109).

In practice, if I have two `watch` commands back to back, I get a deadlock roughly 10% of the time on my Mac.

I have a [somewhat cheesy workaround](https://github.com/michaelfairley/notify/commit/7fd41b3a0a47be4f4013750bd74b0829109db3cd) that I'm using right now, but an ideal solution probably involves `CFRunLoopObserver` or somesuch.

cmyr:
I think this bug is serious enough that the proposed patch should be accepted. I'm not sure it's worth making major changes to `FsEventWatcher` when #117 will probably require another one. @passcod do you have any thoughts?

passcod:
@michaelfairley are you okay with me using your workaround?

michaelfairley:
Yes! Please do! It's been working well enough for me since I initially filed this issues.

cmyr:
@michaelfairley are you in a position to PR your patch? If not I can, save @passcod the time.

passcod:
Done with #134

PR #133, opened 2017-11-22
Make DebouncedEvent be Clone, expose is_recursive

cmyr:
A speculative PR:

These are two very small changes that would be useful to me. Getting #131 onto crates.io would be appreciated, and if you want to include these that would be additionally so; but I do have some workarounds in place already so this is less important.

If you were going to do a bump for crates.io, and could give an up or down vote on #118 first, it could go in at the same time?

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -358,7 +358,7 @@ pub struct RawEvent {
 
 unsafe impl Send for RawEvent {}
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 /// Event delivered when action occurs on a watched path in debounced mode
 pub enum DebouncedEvent {
     /// `NoticeWrite` is emitted immediately after the first write event for the path.


passcod:
The Clone derive is fine, but can you explain your use for the expose?

cmyr:
I have a wrapper API, and I have (perhaps dubiously, I admit) chosen to expose the recursive functionality via a bool, so that consumers don't need to bring `notify::RecursiveMode` into scope. 🤷‍♂️

cmyr:
@passcod I'm backtracking, I think I was being too clever. This is now just the `Clone`.

passcod:
Note for when I cut the release: this is a bump-minor.

michaelfairley:
This doesn't seem to build. 🙁

```
error[E0277]: the trait bound `Error: std::clone::Clone` is not satisfied
   --> src/lib.rs:415:11
    |
415 |     Error(Error, Option<PathBuf>),
    |           ^^^^^^ the trait `std::clone::Clone` is not implemented for `Error`
    |
    = note: required by `std::clone::Clone::clone`
```

Add a derive for `Clone` onto `Error` results in this, which doesn't seem resolvable:
```
error[E0277]: the trait bound `std::io::Error: std::clone::Clone` is not satisfied
   --> src/lib.rs:445:8
    |
445 |     Io(io::Error),
    |        ^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `std::io::Error`
    |
    = note: required by `std::clone::Clone::clone`
```

cmyr:
@michaelfairley ugh good catch. I forgot in my workaround I was converting io::Error into a general error.

I don't suppose that's an appropriate approach here. So:

- We could wrap that error in an Arc
- We could revert this and let me figure it out on my own

Honestly, the latter might be the better option?

cmyr:
@passcod you saw this? I'll put a revert together.

cmyr:
ah beat me to it, great

passcod:
Ah, no, Github's own revert tried to also revert #134, when I just wanted to revert this one. If you can fix before the weekend (when I'll definitely cut a release) don't bother with a revert, though.

passcod:
Hmm, I did the revert. Anyway, you can expect a release this Saturday at the latest.

cmyr:
@passcod great, thanks for this!

Issue #137, opened 2017-12-28
Write more backend compliance tests


passcod:
- [ ] Deep recursive (4+ levels of folders)
- [ ] Wide recursive (lots of shallow folders within top)
- [x] EmitOnAccess
- [x] FollowSymlinks
- [x] renames (TrackRelated)
- [ ] WatchNewFolders
- [ ] Trying to watch a non-existent file/folder

#### Want to help?

Compliance tests live here: https://github.com/passcod/notify/blob/next/backend/src/compliance.rs

Each test:
- should **always pass** if the Backend tested doesn't have the relevant Capability.
- should check events returned from the backend in the **broadest way possible**. E.g. trying to test EmitOnAccess for specifically `Access(Open(Write))` will fail miserably on a backend that only emits `Access(Any)`, even though it should pass.
- should be named clearly. Tests that look at specific Capability should start with `cap_`.
- should be tested on **every Backend on all supported platforms**. This can be done fairly simply by pushing and opening a Pull Request and waiting for CI to ✨ Do Its Thing ✨ 

You can implement one or many of these compliance tests (and be awesome all the same!). Also if you have ideas for things that would make good compliance tests, you can contribute those, too.

passcod:
Closing some v5 issues that are outdated and/or not actionable now.

PR #138, opened 2017-12-28
Add some backend compliance tests

dfaust:
Here are some (incomplete) backend compliance tests for EmitOnAccess, FollowSymlinks and TrackRelated as mentioned in #137.


passcod:
Awesome! 👍 

I'm running the "allowed failures" Travis builds locally (7 through 10) as these only fail on CI and pass just fine locally, to check before I merge.

Issue #139, opened 2017-12-29
Write FSEvent backend


passcod:
- [ ] Implement
- [ ] Test on at least the last two versions of macOS (ref #128 for issues under High Sierra)
- [ ] Document
- [ ] Add to tests
- [ ] Add to Notify (maintainer)

#### Want to help?

As there's no "Backend Implementor's Guide" right now, you'll need to read through raw docs and source to get an idea of what's needed.

Create a new subordinate crate:
- make a `backend-fsevent` folder;
- add a `Cargo.toml` with `name = "notify-backend-fsevent"`;
- add an entry in the root `Cargo.toml` for the new backend;
- put your code in `backend-fsevent/src/lib.rs`;
- add `backend-fsevents/tests/compliance.rs` and populate it as in other backends.

Refer to the Backend documentation in the source (or render it with `cargo doc` yourself for convenience):

- https://github.com/passcod/notify/blob/next/backend/src/backend.rs
- https://github.com/passcod/notify/blob/next/backend/src/lib.rs
- etc

Also look at existing Backends e.g. https://github.com/passcod/notify/blob/next/backend-inotify/src/lib.rs

Read the documentation on `futures::Stream`, especially `poll()` method description of semantics: https://docs.rs/futures/0.1.17/futures/stream/trait.Stream.html#tymethod.poll

When starting out, omit the `#![deny(missing_docs)]` and just get it building and working, first.

You'll need to use the [fsevent](https://github.com/octplane/fsevent-rust) crate.

When you've got something that works, add it to the tests:
- in `ci/cross-tests.sh`
- in `ci/script.sh`
- in `CONTRIBUTING.md`

Then create a Pull Request so the tests will run on Travis!

passcod:
@cmyr

cmyr:
Great, I'll dig into this in the next few days.

passcod:
Thank you so much ❤️

cmyr:
@passcod is there a good way to ping you with little questions I might have? Do you prefer an issue, or irc, or?

passcod:
I'm on the Mozilla IRC, if that works? If I'm not there for some reason (I have a bouncer but who knows) a comment on here is okay.

passcod:
Closing some v5 issues that are outdated and/or not actionable now.

PR #144, opened 2018-01-02
[WIP] FSEvent backend for 5.0

cmyr:
Very much a rough sketch/first-pass right now (I have no idea if this runs/works, as I'm writing this), but wanted to see if I could get CI running. Will tighten this up in coming days.

progress towards #139


backend-fsevent/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    fn caps(&self) -> Vec<Capability> {
+        Self::capabilities()
+    }
+
+    fn capabilities() -> Vec<Capability> {
+        vec![
+            Capability::FollowSymlinks,
+            Capability::WatchEntireFilesystem,
+            Capability::WatchFiles,
```
passcod: I've only given a cursory look to the FSEvent documentation, but it is my understanding that it cannot do that, i.e. FSEvent only watches directories (and [Apple recommends using kqueue for single files instead](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html#//apple_ref/doc/uid/TP40005289-CH5-SW2)).

cmyr: With the `FileEvents` flag available from 10.7, if I create an `FSEventStream` with a single path, to a file, I will receive only events for that specific file, not events in the parent directory.

My impression is that a lot of Apple's documentation on the matter was written for the initial release, and hasn't been updated.

passcod: Oh cool :)

cmyr: Yea, I was pleasantly surprised, because all the docs suggested this wasn't the case. :)

cmyr:
@passcod FSEvents will hold back on sending modify events while the modifying process has an open file descriptor. This means that the  current tests checking for modification events will fail consistently.

More than that: closing the fd does not guarantee that events are flushed; even when you ask the stream not to buffer events, _some_ buffering occurs. To receive multiple events for multiple writes, I have to both close the fd _and_ wait 500ms between them.

More generally, on the filesystem-dev mailing list there's a very clear statement from a developer relations person that FSEvent is _not_ a log of file system activity:

>FSEvents is designed to notify you about events in the file system so that you can then go look in the file system to find the current state of things and sync based on that.  It's not designed to feed you a stream of events with sufficient fidelity to reconstruct how the file system go into its current state."

– [filesystem-dev](https://lists.apple.com/archives/filesystem-dev/2016/Mar/msg00004.html)

So _this particular case_ is not something FSEvent will give us.

It's not totally clear to me how I should handle this. Maybe we want a new capability like "EmitImmediately"? Curious to hear your thoughts.

passcod:
Actually, that's only a problem for the compliance tests, right? It doesn't make a different for usage in Notify or beyond. It's just that this backend will not behave how the compliance tests expect the backends to work.

Instead of adding a capability just to cover this particularity of this backend, we can make an exception to the rule "all backends must pass compliance tests" and write those same tests _for FSEvent_.

So: copy the relevant tests from backend/src/compliance.rs, remove the capability checking, and rewrite them to test this backend with its particularities. When we add more compliance tests we'll need to also add them here, so that's a little more work, but it should work.

cmyr:
Okay that works for me; If you don't think that having the definition of compliance be variable between different backends will be a problem then I am happy to defer to you! I'll have some time tomorrow to work on this, so should at least get the tests done then.

passcod:
@cmyr Hey, have you gotten any further on this? I have my Macbook back now, so I could work on this a bit, but I'd rather not duplicate efforts if you're on it :)

cmyr:
@passcod Sorry I've been travelling a little bit so this fell off my list, but I'll have time early next week at the latest (and probably a few hours later today)

passcod:
No problem!

cmyr:
Okay, I've updated this with a very very basic implementation. We keep track of previously received events; when we receive a new event, we check if there has been a recent event at the same path; if there _has_ we send events for any flags that are in the new event but not the previous; and we also send the `Any` event, to indicate that there _may_ be other changes. This means that if you modify a file twice within, say, 5 seconds, you will receive a `Modify` event followed by an `Any` event; if you create an event and then modify it 5 seconds later, you'll receive a `Create` event, followed by `Modify` _and_ `Any` events.

This doesn't feel great to me, but it's a reasonable starting place?

Pardon the delay, I've been having trouble getting my head back into this.

passcod:
Good work! I like this usage of `Any`. I'll read through more in depth to digest things.

Regarding delay: don't worry :) This isn't blocking anything at the moment — I still have a bunch of work to do on the core interfaces and dayjob has been picking up so I haven't had as much time to work on notify recently. I'm very grateful of any time you do put in this!

Though: if you know — at any point — you're not going to have time for a while (weeks, not days), could you chime in here? That way I can know that I could pick it if I have some time.

cmyr:
My feeling right now is that refining this implementation should probably happen in concert with other work, when you'll be developing a better sense of how you're going to be unifying the various backends, and what would be most helpful for you to that end.

So maybe give this a read when you have time, see if it makes approximate sense, and then it can be something to iterate from. If it looks like it will make sense, I'm happy to add some more documentation.

passcod:
Okay, sounds good. I'll do that.

passcod:
I'll merge this in soonish to work on it in-tree more closely, and bring it up to the newest interface.

passcod:
I looked some more at FSEvent and had a think and changed the plan... rather dramatically. The low down is this: FSEvent is so radically different that it I think it breaks user expectations of what a filesystem notifier should do. This is also supported by the prose in the Apple documentation. Their recommendation for the kind of semantics we're looking for is kqueue.

The plan becomes this:
- FSEvent will become an external notify backend. It might be in this repo, it might not. It won't be included in the "default" notify distribution, and won't be loaded by default. To use it, people will have to require the `notify-backend-fsevent` crate.
- Actual implementation of this backend is pushed back to after Notify beta is out. That is, I remove the dependence on wrangling FSEvent from shipping Notify.
- macOS will be covered with a new kqueue backend, along with BSDs.

That makes the default macOS experience more in line with expectations, at the cost of efficiency (i.e. might see higher CPU usage) when watching large trees. People with requirements where FSEvent fits will be able to opt-in.

I'm closing this PR, but be sure that this work is not wasted by any means! There's a merge-fsevent branch in the repo and that will form the basis for the backend, later on.

cmyr:
I agree that kqueue is likely the correct API for us on macOS. Looking forward to the new work!

Issue #142, opened 2017-12-29
Write tree polling backend


passcod:
I haven't quite figured out the details yet, so no friendly description and no help-wanted tag. Some notes:

Tree-based instead of list-based. Initialisation would build an in-memory tree of the watched path, then subsequent polls would check up on the file structure and emit changes into the Buffer.

The tree structure need not be directly crawled to a tree: we can use `walkdir` to do the actual walking of the tree and re-interpret its result (in the iter loop) into a tree (and compare/update the tree on subsequent walks). That lets `walkdir` handle the pesky details of walking file trees.

- Can detect creates and removes
- Can emit on size change
- Can emit on metadata change
- Can't emit on access
- Can automatically watch new folders
- Can follow symlinks (through `walkdir`)
- No trivial way to track (file) renames
- Tracking moves (path renames) is also non-trivial

The actual walking work (beyond initialisation?) should be done in a subthread (or more) and results pushed back to the thread or written to the Buffer. The `poll()` method would then only read-and-delete from the Buffer.

On some platforms/filesystems, we can perhaps rely on folder timestamps to figure out where not to look further?

> For the far future, adding some heuristic to detect file moves would be great — even beyond poll-watching. An application might be interested to know if files were moved before it starts up a fs watch or even use that information without watching the fs at all. So this should go into a separate crate.

dfaust:
- Currently the fs is read by `walkdir` this has the advantage that some corner cases like symlink loops and file descriptor limits are already handled. When using a tree structure, it won't fit the flat `walkdir` concept anymore. The biggest problem I can see is detecting removes. So we have two options.

1. Implement some of `walkdir`'s functionalities our self and walk the fs recursively. (the walkdir code is relatively small, but of cause the devil is in the detail)
2. Map the flat structure from `walkdir` back to a tree structure. (adds complexity, but should be doable)

- I don't think we should be doing blocking I/O in a poll. `std::fs::read_dir` is blocking and there doesn't seem to be a non-blocking alternative. Therefore I would suggest doing that in a thread. This would emulate the behavior of the other back-ends and would lead to better predictability.

- For the far future, adding some heuristic to detect file moves would be great - even beyond poll-watching. An application might be interested to know if files were moved before it starts up a fs watch or even use that information without watching the fs at all. So this should go into a separate crate.

passcod:
Also could do a port of the existing flat polling backend to start with.

passcod:
Closing some v5 issues that are outdated and/or not actionable now.

PR #145, opened 2018-01-03
[WIP] Implement poll watcher

dfaust:
I started implementing the poll watcher. There's still a lot missing, but it passes all existing tests.

As discussed in #142, the code is quite ugly. In addition to a tree, that holds all files and directories, there is an `ancestors` vector that holds the children of all parent directories while traversing the fs using walkdir. It will be used to detect deletes.

There are still some bigger problems:
- The poll watcher needs an `interval` parameter. It's currently set to 20ms so the tests can succeed.
- Stopping the poll thread is problematic. It currently checks for a shutdown signal once every iteration of the main loop. This is too slow, especially if the `interval` is very high.


passcod:
For the second issue, how about this:

- Instead of `sleep`, use `park_timeout_ms`
- Store the thread Handle on the Backend
- When the Backend Drops, tell the channel to shutdown, then unpark the thread

That should get rid of the wait without having a tiny loop interval.

SirVer:
@dfaust Thanks for putting so much work in improving notify! Félix made it excellent already, but your redesign looks super clean right now and I am excited about all the new features!

dfaust:
Well, as you already noticed, I didn't do the re-design, that's all @passcod's work.

passcod:
I'll merge this soonish to bring it within the main tree and bring it to the newest interface, which I believe might neatly solve some of its issues.

PR #151, opened 2018-02-18
Use tokio to drive the event streams in the compliance tests

dfaust:
In order for this to work, the backends must notify the tokio executor when the stream should be polled again.

I already implemented executor notifications for the poll backend, but not for the others. I haven't researched how that could be done, yet, but it might not be that easy.


passcod:
Oh that is much better. I've also been looking at the executor notifications and the new tokio APIs... I believe it should not be very difficult to implement for other backends, as that's much closer to how they expect to be used.

I've been playing with tokio in a side project to get a better feel for how it all works and the ergonomics as a client/consumer rather than a producer. Keeping in mind that futures/tokio 0.2 are coming in 6-8 weeks with more breaking changes, though...

PR #156, opened 2018-04-07
Refer to macOS by that name rather than OS X.

waywardmonkeys:


passcod:
:+1:

[non-breaking, patch]

PR #159, opened 2018-08-05
Poll backend: Release the lock while we sleep.

Freaky:
If we sleep while holding the lock, we can block calls to watch().

Seeing this on FreeBSD, where even the most trivial example program will randomly hang while setting up watches.


Keats:
Looks like the OSX errors are legit?

Freaky:
They don't look related to this commit, same failures are seen in the [last master CI run](https://travis-ci.org/passcod/notify/builds/366433646).

PR #162, opened 2018-08-25
Update dependencies

dimbleby:
I've updated a bunch of dependencies, in a series of commits.

Much the biggest was `mio`.  

I don't have access to a mac, so `fsevent` is left as an exercise for the reader.


passcod:
Wow, good work! Thank you very much

I think this is alright, but I'll make CI run a few more times to get the spurious failures some extra coverage before merging.

passcod:
Hmm, the windows test fails consistently (not just in this PR) on `move_in_directory_watch_subdirectories`... would you have any insight given you looked at that code? Does it fail when run locally, or is it one of those pesky "fails only in CI" things?

Also could you move `inotify/mod.rs` to `inotify.rs` given the flags file is gone now :)

dimbleby:
Windows tests failed for me consistently before I started - all of `create_directory_watch_subdirectories`, `create_rename_overwrite_file` and `move_in_directory_watch_subdirectories` fail locally from master.  I'm afraid I haven't investigated, sorry.

I'll move the code around as you propose and push shortly.

I'd suggest not merging this - or certainly not publishing it - until a fix for that `inotify` bug is released; and then presumably we'll want to depend on version 0.6.1 rather than 0.6.0.

Cargo.toml
```diff
@@ -34,7 +34,7 @@ fsevent = "^0.2.17"
 fsevent-sys = "^0.1.3"
 
 [target.'cfg(windows)'.dependencies]
-winapi = "^0.2.5"
+winapi = "^0.3.5"
 kernel32-sys = "^0.2.1"
```
crlf0710: When upgrading winapi to "0.3", `kernel32-sys` should be dropped.

dimbleby: ack, thanks

hannobraun:
I've merged https://github.com/inotify-rs/inotify/pull/117 and released inotify 0.6.1.

dimbleby:
@hannobraun thanks!  I've updated this pull request to match

passcod:
Thanks all! I'll merge and release tomorrow (for me). In ~12 hours

passcod:
Released in 4.0.5

PR #166, opened 2018-09-29
Explain how watch does not follow moves or renames

devurandom:
This summarises the knowledge gathered from https://github.com/passcod/notify/issues/165#issuecomment-425631193

Closes: https://github.com/passcod/notify/issues/165


passcod:
Well, yes, but this is only correct for Linux and macOS.

- Polling doesn't follow (it can't), but will pick up later if something is moved onto the path it's watching (it doesn't know it's not the original).
- On Windows, the platform API gets given a handle to a directory, not a path. It's somewhat unclear if that handle remains the same after a move, but I would guess it does.

PR #170, opened 2019-01-12
fix for non terminanting debounced watcher

### LINKED ISSUE #169: debounced watcher doesn't terminate
This code doesn't terminate: https://github.com/vemoo/notify/blob/a02645e73eb4b0ec3de5ae7774b2a59cf4dc33ca/tests/debounce.rs#L1117-L1125

The issue seems to be with the `WatchTimer` thread not terminating.
I've changed the code here https://github.com/vemoo/notify/commit/a02645e73eb4b0ec3de5ae7774b2a59cf4dc33ca and that way it terminates, but I'm not sure if there's a better way.

vemoo:
fixes #169

### DIFF
--- a/src/debounce/timer.rs
+++ b/src/debounce/timer.rs
@@ -13,6 +13,7 @@ use debounce::OperationsBuffer;
 enum Action {
     Schedule(ScheduledEvent),
     Ignore(u64),
+    Stop
 }
 
 #[derive(PartialEq, Eq)]
@@ -73,8 +74,9 @@ impl ScheduleWorker {
                         }
                     }
                 },
-                Err(mpsc::TryRecvError::Empty)=> break,
-                Err(mpsc::TryRecvError::Disconnected)=> {
+                Err(mpsc::TryRecvError::Empty) => break,
+                Err(mpsc::TryRecvError::Disconnected)
+                | Ok(Action::Stop) => {
                     self.stopped = true;
                     break;
                 },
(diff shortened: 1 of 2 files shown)


src/debounce/timer.rs
```diff
@@ -70,9 +71,11 @@ impl ScheduleWorker {
                             break;
                         }
                     }
-                }
+                },
+                Action::Stop => return false
             }
         }
+        true
```
matklad: Have *zero* idea about what's going on here, but usually channel's closure signifies the same thing as an explicit stop, so I'd expect to see `false` here as well.

vemoo: But you can end up there if the channel is empty.
But now that think about it I could have just handled the error case explicitly and return `false` for `Disconnected` and true for `Empty`.
Then on `drop` for `WatchTimer` I just have to drop the sender before the `self.trigger.notify_one();`

matklad: >Then on drop for WatchTimer I just have to drop the sender before the 

Again, don't really understand the context, but this seems right to me: it's almost always possible to use "receiver is closed when sender is dropped" instead of an explicit `Stop` event. The benefit is that, after dropping the sender, you have to way to send additional messages.

vemoo: It seems it doesn't work, because the `Sender` can't be dropped before `self.trigger.notify_one();` and that needs to happen to avoid a race condition.

Issue #173, opened 2019-01-21
inotify backend: frequent wakeups every 100ms


WGH-:
Not a good thing for powersaving. The problematic parts are here:

https://github.com/passcod/notify/blob/6e211f79b2c3480ea47f198d46a1b9c326e259f1/src/inotify.rs#L149-L165

passcod:
Released in 4.0.7.

PR #174, opened 2019-01-22
Remove timeout on inotify poll

passcod:
This was added in #162 but I don't think it's actually needed. All events go through this, even shutdown events (so we don't need to handle signals ourselves / use `poll_interruptible`).

Reported in #173


passcod:
[patch bump]

PR #195, opened 2019-05-06
Update fsevent dependencies

Eijebong:
This will dedupe quite a few dependencies in crates using notify. I didn't bump the version because I don't know if any breaking change mad it to the branch but I'd be more than happy to do so if you want

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,8 @@ mio = "^0.6.15"
 mio-extras = "^2.0.5"
 
 [target.'cfg(target_os="macos")'.dependencies]
-fsevent = "^0.2.17"
-fsevent-sys = "^0.1.3"
+fsevent = "0.4"
+fsevent-sys = "2"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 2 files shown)


passcod:
Is this a WIP? It doesn't build at all.

Eijebong:
Weird, it built for me on linux and macos, I'll have a look

Eijebong:
Ok, that's funy, it doesn't fail on macos but it fails if I cross compile from linux... i'll fix this

passcod:
Does the test suite fail for you locally? I'll retry the run in CI to see if it's intermittent anyway.

Otherwise, looks good. That will get into 5.0.

For the current major, would you be willing to also write this for the `try-v4` branch I just pushed? That way I'd do a patch release immediately.

Eijebong:
This looks intermittent. Was it already like that before ? Maybe some time behavior changed and needs adjustment ?

passcod:
Fsevents tests are awful and often intermittent, this is known :)

Eijebong:
Ok, I'll make a PR against the `try-v4 branch` too then

PR #196, opened 2019-05-06
Update fsevent dependencies and bump version

Eijebong:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "notify"
-version = "4.0.10"
+version = "4.0.11"
 authors = [
   "Félix Saparelli <me@passcod.name>",
   "Jorge Israel Peña <jorge.israel.p@gmail.com>",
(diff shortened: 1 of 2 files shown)


passcod:
Ah, this build shows some failures on 1.26.1. Some of it looks like module/paths, should be pretty simple:

```
error[E0433]: failed to resolve. Use of undeclared type or module `std`
   --> src/fsevent.rs:130:43
    |
130 |             let mut err: cf::CFErrorRef = std::ptr::null_mut();
    |                                           ^^^ Use of undeclared type or module `std`
```

And some of it is type mismatches, and I'm not super sure what's going on there:

```
error[E0308]: mismatched types
   --> src/fsevent.rs:107:43
    |
107 |                 while !CFRunLoopIsWaiting(runloop) {
    |                                           ^^^^^^^ expected enum `std::os::raw::c_void`, found enum `libc::c_void`
    |
    = note: expected type `*mut std::os::raw::c_void`
               found type `*mut libc::c_void`
```

Perhaps some stdlib changes?

Eijebong:
Should be fixed now. I couldn't run the tests but it compiled

PR #198, opened 2019-05-21
impl Sync for PollWatcher

rabite0:
After adding cross compiled binaries for hunter using travis I noticed that it failed building on FreeBSD because PollWatcher doesn't implement Sync. Since all methods take &mut self, I guess this is OK.

Same thing, just for v4.


passcod:
Can you fix for 1.26.1? See the travis log

Issue #205, opened 2019-06-14
Thread watching debounced events hanging


itkovian:
#### System details

- Linux, CentOS 7.6, kernel 3.10.0-957.12.2.el7.ug.x86_64 (custom rebuild, with some support for GPFS)
- rustc 1.34.2 (6c2484dc3 2019-05-13)
- notify main brach, commit 56aac12

I use notify in https://github.com/itkovian/sarchive/. The issue I opened for this is at https://github.com/itkovian/sarchive/issues/20. 

- FS: /dev/mapper/vg0-scratch on /tmp type ext4 (rw,relatime,data=ordered)
- Hypervisor is kvm, in an OpenNebula env. 
- Underlying storage is Ceph (but I do not think that is relevant here)

#### What you did (as detailed as you can)

- I created 50K files, in a directory that was watched. The watcher threads always stops, but not always after the same number of files. 
- The notify queue size was raised to 64K
- These are the events seen:
~~~~
[2019-06-14T11:36:08.930803788+02:00][sarchive::lib][DEBUG] Event received: Event { kind: Create(Any), paths: ["/tmp/torque/1.SC"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
[2019-06-14T11:36:08.931012465+02:00][sarchive::lib][DEBUG] Event received: Event { kind: Create(Any), paths: ["/tmp/torque/2.SC"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
[2019-06-14T11:36:08.931178128+02:00][sarchive::lib][DEBUG] Event received: Event { kind: Create(Any), paths: ["/tmp/torque/3.SC"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
[2019-06-14T11:36:08.931341658+02:00][sarchive::lib][DEBUG] Event received: Event { kind: Create(Any), paths: ["/tmp/torque/4.SC"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
[2019-06-14T11:36:08.931504147+02:00][sarchive::lib][DEBUG] Event received: Event { kind: Create(Any), paths: ["/tmp/torque/5.SC"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
[...]

PR #210, opened 2019-09-11
fix deadlock in debouncer

mtak-:
Works around the issue described in #205. Tested on the ~fsevent backend only~ linux/macos/windows.

I tried to change locking as little as possible as it's somewhat hard to follow, and instead use a simple backoff algorithm to workaround the deadlock.

These tests fail on both branch `origin/try-v4` and my fork.
```
test create_rename_overwrite_directory ... FAILED
test create_rename_overwrite_file ... FAILED
```

Would be nice to get a new version out with the fix as well!

### DIFF
--- a/src/debounce/mod.rs
+++ b/src/debounce/mod.rs
@@ -12,8 +12,8 @@ use std::sync::mpsc;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-pub type OperationsBuffer =
-    Arc<Mutex<HashMap<PathBuf, (Option<op::Op>, Option<PathBuf>, Option<u64>)>>>;
+pub type OperationsBufferInner = HashMap<PathBuf, (Option<op::Op>, Option<PathBuf>, Option<u64>)>;
+pub type OperationsBuffer = Arc<Mutex<OperationsBufferInner>>;
 
 pub enum EventTx {
     Raw {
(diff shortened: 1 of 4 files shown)


mtak-:
Many of the ci failures existed previously: https://travis-ci.org/passcod/notify/builds/579460084

0xpr03:
I'll have to check this when I've got some time. Please ping me again otherwise in 3 days.

0xpr03:
I think we want to warn users about the broken state of the debouncer in these cases, if we're not able to fix it sufficiently. This affects at least 4.X stable and 5.X. So breaking the debouncer isn't possible. Your PR also mentions macOS specifically. This issue persists at least also on linux, it may also be another issue with similar symptoms.
cc @Dylan-DPC

mtak-:
The `delay_zero` test repros the issue pretty consistently on macos. If someone wants to test linux, just pasting that test into the current version should hopefully fail. Whereas running that test against this PR should hopefully succeed.

mtak-:
I pushed up the `delay_zero` test _without_ the deadlock fix workaround, to let travis test the platforms I don't have access to (windows/linux). The test failed on both platforms [here](https://travis-ci.com/notify-rs/notify/builds/129158452). Altho the `delay_zero` test did pass one of the linux builds (it is a non-deterministic test).

Testing both platforms again with the deadlock fix shows the `delay_zero` test passing on all platforms [here](https://travis-ci.com/notify-rs/notify/jobs/239151520).

0xpr03:
Thanks for all the work. I re-scheduled CI and will verify on linux again.

0xpr03:
I can confirm that this resolves the deadlock problem under linux.
I can't say why `truncate_write_file` failed on windows for 1.26 but would merge this. cc @JohnTitor

PR #214, opened 2019-09-16
Remove channel in favour of accepting `Fn` in Watcher constructors

### LINKED ISSUE #160: Use crossbeam-channel instead of stdlib
Hi! My understanding is that currently `notify` hard-codes the usage of `std::mpsc`, and this is a bit unfortunate: there are many channel implementations in Rust (`crossbeam-channel` most notably), and the std version is not really great: it lacks `select`, which severally  restricts its applicability. 

It would be cool if `notify` had a `trait EventSink { fn send(&mut self, Event); }` and used that instead of any particular channel implementation. 

Though, for practical purposes, perhaps just betting on `crossbeam-channel` would be even better?

### LINKED ISSUE #213: Have `Watcher::new` accept a function rather than a channel
## Motivation

It appears the current channel-based watcher API might be a little unnecessarily opinionated, as it requires users not only use a queue for handling their events, but a specific channel type from a specific crate (previously std, now crossbeam-channel). This adds a little indirection if users simply want to update something behind a `Mutex`, or perhaps print something to stdio, etc. It also requires adding an extra branch of dependencies (via crossbeam-channel) when the user may not require them otherwise at all. Finally, further indirection is necessary if the user wishes to select across a channel type from another package, perhaps to fit into some larger existing async project.

## API Change Proposal

I propose changing the `Watcher::new` / `Watcher::new_immediate` APIs to accept a function of type `F: Fn(Result<Event>)` rather than a `Sender` half of a channel. This still allows the user to use a channel if they wish e.g.
```rust
let w = RecommendedWatcher::new_immediate(move |res| tx.send(res)).unwrap();
```
but also allows a little more flexibility in the case the user doesn't require a queue at all. e.g.

```rust
let w = RecommendedWatcher::new_immediate(move |res| {
    if let Ok(ev) = res {
        model.lock().unwrap().update_by_event(ev);
    }
}).unwrap();
```

Doing something like this with the current API is of course possible today, but currently requires either blocking the current thread or spawning a new one along with the aforementioned issues related to requiring a specific kind of channel.

Note the actual bounds required by the function will be more like:
```rust
F: 'static + Fn(Result<Event>) + Send,
```
as I'm guessing there may be a requirement to move or share the function between one or more threads within the individual watcher implementations. This constraint could be made a little more concise with a trait along the lines of:

```rust
pub trait EventFn: 'static + Fn(Result<Event>) + Send {}
impl<F> EventFn for F where F: 'static + Fn(Result<Event>) + Send {}
```

mitchmindtree:
This implements the changes described in #213.

~~This PR is a **work-in-progress**.~~

TODO:

- [x] Make API change.
- [x] Update inotify watcher.
- [x] Update poll watcher.
- [x] Update fsevents watcher.
- [x] Update windows watcher.

Closes #213.
Closes #160.

### DIFF
--- a/examples/monitor_raw.rs
+++ b/examples/monitor_raw.rs
@@ -2,20 +2,24 @@ use notify::{RecommendedWatcher, RecursiveMode, Watcher};
 use std::path::Path;
 
 fn watch<P: AsRef<Path>>(path: P) -> notify::Result<()> {
+    let (tx, rx) = std::sync::mpsc::channel();
+
     // Automatically select the best implementation for your platform.
     // You can also access each implementation directly e.g. INotifyWatcher.
-    let mut watcher: RecommendedWatcher = Watcher::new_immediate(|res| {
-        match res {
-            Ok(event) => println!("changed: {:?}", event),
-            Err(e) => println!("watch error: {:?}", e),
-        }
-    })?;
+    let mut watcher: RecommendedWatcher = Watcher::new_immediate(move |res| tx.send(res).unwrap())?;
 
     // Add a path to be watched. All files and directories at that path and
     // below will be monitored for changes.
     watcher.watch(path, RecursiveMode::Recursive)?;
 
-    loop {}
+    for res in rx {
+        match res {
+            Ok(event) => println!("changed: {:?}", event),
+            Err(e) => println!("watch error: {:?}", e),
+        }
+    }
+
+    Ok(())
 }
 
 fn main() {
--- a/examples/monitor_raw.rs
+++ b/examples/monitor_raw.rs
@@ -1,27 +1,21 @@
-use crossbeam_channel::unbounded;
 use notify::{RecommendedWatcher, RecursiveMode, Watcher};
 use std::path::Path;
 
 fn watch<P: AsRef<Path>>(path: P) -> notify::Result<()> {
-    // Create a channel to receive the events.
-    let (tx, rx) = unbounded();
-
     // Automatically select the best implementation for your platform.
     // You can also access each implementation directly e.g. INotifyWatcher.
-    let mut watcher: RecommendedWatcher = Watcher::new_immediate(tx)?;
+    let mut watcher: RecommendedWatcher = Watcher::new_immediate(|res| {
+        match res {
+            Ok(event) => println!("changed: {:?}", event),
+            Err(e) => println!("watch error: {:?}", e),
+        }
+    })?;
 
     // Add a path to be watched. All files and directories at that path and
     // below will be monitored for changes.
     watcher.watch(path, RecursiveMode::Recursive)?;
 
-    // This is a simple loop, but you may want to use more complex logic here,
-    // for example to handle I/O.
-    loop {
-        match rx.recv() {
-            Ok(event) => println!("changed: {:?}", event),
-            Err(e) => println!("watch error: {:?}", e),
-        }
-    }
+    loop {}
 }
 
 fn main() {
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -14,7 +14,7 @@
 
 #![allow(non_upper_case_globals, dead_code)]
 
-use crate::{Config, Error, EventTx, RecursiveMode, Result, Watcher};
+use crate::{Config, Error, EventFn, RecursiveMode, Result, Watcher};
 use crate::event::*;
 use crossbeam_channel::{unbounded, Receiver, Sender};
 use fsevent as fse;
@@ -30,6 +30,7 @@ use std::path::{Path, PathBuf};
 use std::ptr;
 use std::slice;
 use std::str::from_utf8;
+use std::sync::Arc;
 use std::thread;
 
 /// FSEvents-based `Watcher` implementation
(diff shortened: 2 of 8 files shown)


mitchmindtree:
Oh also, I don't have access to macOS or Windows at the moment - if anyone has access to either of those platforms and would like to include them in this PR feel free! Otherwise it might be a while until I can address this.

mitchmindtree:
OK, I've managed to update the `windows` and `fsevents` `Watcher` implementations by adding windows and darwin targets via rustup and `cargo check`ing those targets locally - I think this should be ready to go!

mitchmindtree:
@matklad ah I just noticed you requested a change similar to this in #160 for rust-analyser. Would you be interested in reviewing this?

examples/monitor_raw.rs
```diff
@@ -1,27 +1,21 @@
...
 
-    // This is a simple loop, but you may want to use more complex logic here,
-    // for example to handle I/O.
-    loop {
-        match rx.recv() {
-            Ok(event) => println!("changed: {:?}", event),
-            Err(e) => println!("watch error: {:?}", e),
-        }
-    }
+    loop {}
```
matklad: This looks very odd. Let's just put an event into a channel in the callback?

src/fsevent.rs
```diff
@@ -485,15 +493,5 @@ fn test_fsevent_watcher_drop() {
...
 
-    // if drop() works, this loop will quit after all Sender freed
-    // otherwise will block forever
-    for e in rx.iter() {
-        println!(
-            "debug => {:?} {:?}",
-            e.op.map(|e| e.bits()).unwrap_or(0),
-            e.path
-        );
-    }
```
matklad: Again, here we'd want to send a message into a channel to properly synchronize

src/inotify.rs
```diff
@@ -517,9 +516,9 @@ fn filter_dir(e: walkdir::Result<walkdir::DirEntry>) -> Option<walkdir::DirEntry
 }
 
 impl Watcher for INotifyWatcher {
-    fn new_immediate(tx: Sender<Result<Event>>) -> Result<INotifyWatcher> {
+    fn new_immediate<F: EventFn>(event_fn: F) -> Result<INotifyWatcher> {
```
matklad: I'd use `Box<dyn EventFn>` here, to avoid extra monomorphisations.

src/lib.rs
```diff
@@ -130,6 +126,13 @@ pub mod poll;
 mod config;
 mod error;
 
+/// The set of requirements for watcher event handling functions.
+pub trait EventFn: 'static + Fn(Result<Event>) + Send + Sync {}
+
+impl<F> EventFn for F
+where
+    F: 'static + Fn(Result<Event>) + Send + Sync {}
```
matklad: I'd use something like `type OnEvent = dyn Fn(Result<Event>) + Send + Sync + 'static`

src/lib.rs
```diff
@@ -139,7 +142,9 @@ pub trait Watcher: Sized {
     /// Create a new watcher in _immediate_ mode.
     ///
     /// Events will be sent using the provided `tx` immediately after they occur.
-    fn new_immediate(tx: Sender<Result<Event>>) -> Result<Self>;
+    fn new_immediate<F>(event_fn: F) -> Result<Self>
+    where
+        F: EventFn;
```
matklad: Heh, on one hand, making this generic doesn't make sense, as we use dynamic dipatch internally anyway. On the other hand, generics give us a nicer syntax, where you can pass-in the lambda directly. I don't have a specific advice here, but I'd probably try to arrange API in such a way that publich API is generic, but immediately boxes the closure into an Arc, and calls the private impl

src/poll.rs
```diff
@@ -197,8 +187,8 @@ impl PollWatcher {
 }
 
 impl Watcher for PollWatcher {
-    fn new_immediate(tx: Sender<Result<Event>>) -> Result<PollWatcher> {
-        PollWatcher::with_delay(tx, Duration::from_secs(30))
+    fn new_immediate<F: EventFn>(event_fn: F) -> Result<PollWatcher> {
+        PollWatcher::with_delay(event_fn, Duration::from_secs(30))
     }
 
     fn watch<P: AsRef<Path>>(&mut self, path: P, recursive_mode: RecursiveMode) -> Result<()> {
```
matklad: Unrelated, but this suffers from the same problem: it's a large generic function which is monomorphised for every `P`.

src/lib.rs
mitchmindtree: Haha yes, I have been conscious of this in each of the backends, that is each of the implementations almost immediately call some function into which the generic function is either `Box`d or `Arc`d depending on the needs of the backend. I figure this way we get the dynamic dispatch indirection to avoid large monomorphisation while still getting that nice generic API and only a single point of dynamic dispatch indirection. You are right about the `PollWatcher` backend though! I just pushed a commit that addresses that backend by having the `PollWatcher::with_delay` function accept an `Arc<dyn EventFn>` instead.

I'll work on addressing some of your other comments soon - thanks for taking a look!

src/lib.rs
mitchmindtree: Thanks for the tip, but does this matter much? Are there bigger motivations for using the type alias, or is it mostly just to avoid having to prefix `dyn` everywhere? I'm currently leaning towards leaving this as it is, but want to check I'm not missing some subtly incurred cost by leaving it this way.

mitchmindtree:
Thanks again @matklad, I think I've addressed everything you have raised.

If there is nothing else, I might merge this within the next couple of days.

A nice follow-up to this PR might be to see if crossbeam is necessary at all any longer. It maybe be possible to remove the remaining use cases in favour of std in order to shave off a dependency?

src/lib.rs
matklad: It’s not to avoid dyn, it’s to avoid an extra trait and just lean on existing Fn

PR #215, opened 2019-09-17
Fix build failures in travis due to filetime

0xpr03:
Should resolve most of the issues with builds on v4


0xpr03:
We will need another PR to fix windows builds.

PR #217, opened 2019-09-21
Dependencies: Update winapi and remove kernel32-sys

rep-nop:
This removes the dependency on `kernel32-sys`, replacing the `kernel32-sys` calls with the appropriate `winapi` calls.

Some neat build time graphs (provided by @memoryruins since I'm on Linux, thanks!)

Before: https://gistpreview.github.io/?c6d113198e6c82d550ffcee0bb156dc6

After: http://gistpreview.github.io/?dcb15cd72b87e4e2ebebea14e360ba7d

This should reduce the build times fairly significantly (shaves off ~14 seconds in those reports) since it removes the old `winapi-0.2.5` dependency brought in by `kernel32-sys`

This has even greater savings for crates that use `notify` itself, link to relevant issue [here](https://github.com/EmbarkStudios/rust-ecosystem/issues/42), specifically [this comment](https://github.com/EmbarkStudios/rust-ecosystem/issues/42#issuecomment-533822802)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,7 @@ fsevent = "1.0.1"
 fsevent-sys = "2"
 
 [target.'cfg(windows)'.dependencies]
-kernel32-sys = "0.2.1"
-winapi = "0.3.5"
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
Thanks

PR #220, opened 2019-09-30
Fix a few URLs

atouchet:


0xpr03:
Look good to me, travis url works again. Anything else ?

passcod:
Probably want to remove the abandonment notice

atouchet:
I think removing the abandonment notice could be done in a followup PR. There are a few other changes that could be made there that I think it would be best for someone else to do.

PR #222, opened 2019-10-01
Remove abandoned description on README

JohnTitor:
Closes #221

PR #224, opened 2019-10-02
Update changelog about maintainers

JohnTitor:
Suggested in https://github.com/notify-rs/notify/pull/222#issuecomment-537213120


JohnTitor:
Failed in installing windows-sdk, is it needed nowadays? I tested on my fork, the CI was passed without it.
CC: @passcod

passcod:
Yeah, not sure. It seems to be a temporary failure on Travis CI, given it was working two days ago. You could try removing it and see

JohnTitor:
Yes, the failure is temporary and the fix is a stopgap but it's worth to remove the installation if it isn't needed. I'll try to do.

JohnTitor:
Opened #225

PR #225, opened 2019-10-02
Use windows-sdk-10.1 to avoid installation failure

JohnTitor:
Install windows-sdk-10.1 instead of 10.0 to avoid installation failure.


JohnTitor:
It's my lack of attention, and easy to fix. We can use `windows-sdk-10.1` to avoid failure.
This doesn't affect codebase so I'm going to merge!

PR #227, opened 2019-10-11
Run rustfmt

JohnTitor:


dhiemaz:
> _No description provided._

@JohnTitor what is this changes is about ?

JohnTitor:
@dhiemaz run `cargo fmt` to format code

0xpr03:
We may want to enforce rust fmt checking during CI

dhiemaz:
> We may want to enforce rust fmt checking during CI

That would be better @0xpr03, any other concern @Dylan-DPC , @passcod ?

passcod:
My experience is generally that running style checking during CI is a deterrent for contributions. While it's not super easy with Travis (but I think there's a ready-made recipe for github actions somewhere), having something auto-commit the result of rustfmt onto PRs (or at merge time) is a lot friendlier.

JohnTitor:
Running rustfmt on CI check is hard work a bit for any contributors even though it keeps code style always. I prefer to submit PR for rustfmt sometimes rather than asking contributors to run rustfmt often.

JohnTitor:
Anyway, I'm going to merge this. If we still want to run rustfmt on CI, please open an issue.

PR #231, opened 2019-11-23
Backport some updates and bump version to 4.0.15

atouchet:


atouchet:
This looks like it still has some issues.

PR #236, opened 2019-12-25
Tweak test since the issue seems resolved

JohnTitor:
It seems the issue on Windows has been gone.


JohnTitor:
Note: the failure on macOS appears randomly

PR #237, opened 2020-01-02
Update dependencies and do some cleanup

JohnTitor:


.travis.yml
```diff
@@ -17,6 +17,8 @@ stages:
   - check nightlies
 
 jobs:
+  allow_failures:
+    - os: osx
```
0xpr03: Feels bad, but I guess it's easier..

JohnTitor: Ah, it was for my debugging, dropped. But actually, if a test fails on the way, stage builds don't run further stages. It's not useful a bit.

PR #256, opened 2020-07-14
Do not stat a path if not recursive.

stuhood:
Changing the order of the checks in `EventLoop::add_watch` avoids a stat for non-recursive watches.

### DIFF
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -404,9 +404,7 @@ impl EventLoop {
     }
 
     fn add_watch(&mut self, path: PathBuf, is_recursive: bool, mut watch_self: bool) -> Result<()> {
-        let metadata = metadata(&path).map_err(Error::io)?;
-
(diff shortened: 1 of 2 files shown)


src/inotify.rs
```diff
@@ -404,9 +404,7 @@ impl EventLoop {
     }
 
     fn add_watch(&mut self, path: PathBuf, is_recursive: bool, mut watch_self: bool) -> Result<()> {
-        let metadata = metadata(&path).map_err(Error::io)?;
-
-        if !metadata.is_dir() || !is_recursive {
+        if !is_recursive || !metadata(&path).map_err(Error::io)?.is_dir() {
```
JohnTitor, line 407: It'd be great if we could add a comment here so that we don't lose this order in the future. Could you also add an entry that mentions this change on the changelog?

PR #260, opened 2020-10-03
Fix typo in documentation link in lib.rs

dwnusbaum:
While browsing through the docs [here](https://docs.rs/notify/5.0.0-pre.3/notify/), I noticed a broken link pointing to https://docs.rs/notify/5.0.0-pre.3/notify/event/index.html%60 caused by a stray backtick in the link target.


JohnTitor (approved):
Good catch, thanks!

PR #268, opened 2020-11-04
EventLoop::new: Use level-triggered events for inotify fd.

### LINKED ISSUE #267: Notify fails to report file creation events if they happen too quickly
When watching a directory, if many files are created in rapid succession, notify reports the first few creations but then stops reporting the rest. If, after a pause, another file is created in the directory, the missing reports do come in.

Linux, Fedora Core 32, x86_64
rustc 1.47.0 (18bf6b4f0 2020-10-07)
notify 5.0.0-pre.4

Test case:
```
use std::fs;
use notify::{Event, EventKind, RecommendedWatcher, RecursiveMode, Result, Watcher};

fn main() -> Result<()> {
    let mut watcher: RecommendedWatcher = Watcher::new_immediate(|res| {
        match res {
            Ok(Event { kind: EventKind::Create(_), paths, .. }) => {
                eprintln!("created {:?}", paths);
            }
            _ => (),
        }
    })?;

    fs::create_dir("notify-hang-test")?;
    watcher.watch("notify-hang-test", RecursiveMode::NonRecursive)?;

    for i in 0..30 {
        fs::File::create(format!("notify-hang-test/glurf{:02}", i))?;
    }

    // Only the first dozen or two files get reported.
    std::thread::sleep(std::time::Duration::from_secs(5));

    // Poking the directory again does seem to wake it up.
    eprintln!("poking");
    fs::File::create("notify-hang-test/glurf31")?;

    // Wait for reports to come in
    std::thread::sleep(std::time::Duration::from_secs(5));

    Ok(())
}
```

Sample output:
```
$ rm -r notify-hang-test
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/play-notify-hang`
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf00"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf01"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf02"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf03"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf04"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf05"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf06"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf07"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf08"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf09"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf10"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf11"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf12"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf13"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf14"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf15"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf16"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf17"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf18"]
poking
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf19"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf20"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf21"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf22"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf23"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf24"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf25"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf26"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf27"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf28"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf29"]
created ["/home/jimb/rust/play-notify-hang/notify-hang-test/glurf31"]
```

jimblandy:
The `inotify` crate's `Inotify::read_events` method does not read all available
events from inotify (see hannobraun/inotify#156), only one buffer's worth. Using
level-triggered events tells Mio to report events on the inotify fd until all
events have been read.

Fixes #267.

### DIFF
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -108,7 +108,10 @@ impl EventLoop {
             &evented_inotify,
             INOTIFY,
             mio::Ready::readable(),
-            mio::PollOpt::edge(),
+            // Use level-sensitive polling (issue #267): `Inotify::read_events`
+            // only consumes one buffer's worth of events at a time, so events
+            // may remain in the inotify fd after calling handle_inotify.
+            mio::PollOpt::level(),
         )?;
 
         let event_loop = EventLoop {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 - CHANGE: Remove `Sender`s from watcher API in favour of `EventFn` [#214]
 - META: The project maintainers have changed from @passcod to notify-rs.
 - CHANGE: Avoid stating the watched path for non-recursive watches with inotify [#256]
+- FIX: Report events promptly on Linux, even when many occur in rapid succession. [#268]
+
+[#268]: https://github.com/notify-rs/notify/pull/268
 
 ## 5.0.0-pre.4 (2020-10-31)
 


JohnTitor:
The solution looks reasonable to me. This may have a negative impact on performance but I think it's a trade-off. @notify-rs/watchmakers Any concerns?

@jimblandy And this bugfix (i.e. behavior change) would be worth mentioning on the changelog, could you add an entry?

jimblandy:
Okay, great! Yes, I'll add a changelog entry.

When the number of events is small enough to be handled by a single call to `read_events`, this shouldn't affect performance at all, unless I'm missing something. When we have so many events that we need multiple `read_events` calls to handle them all, there will be additional trips through the Mio event loop. But I don't think that's especially expensive. We could amortize it further by using a larger buffer.

A more complex fix that minimizes Mio events would be to have `handle_inotify` call `read_events` repeatedly until it returns an iterator that produces no `Event`s. Then we could keep the file descriptor edge-triggered. But I doubt the costs justify that, and I don't want to implement it myself.

jimblandy:
@JohnTitor Would it be worth backporting this to the last version `4` release? That has the same bug.

JohnTitor:
> When the number of events is small enough to be handled by a single call to read_events, this shouldn't affect performance at all, unless I'm missing something. When we have so many events that we need multiple read_events calls to handle them all, there will be additional trips through the Mio event loop. But I don't think that's especially expensive. We could amortize it further by using a larger buffer.

Makes sense, then there's no roadblock to merge!

> Would it be worth backporting this to the last version 4 release? That has the same bug.

I believe it's safe to backport, yes. If you want, I'm happy to accept it :)

JohnTitor (approved):

src/inotify.rs
```diff
@@ -108,7 +108,7 @@ impl EventLoop {
             &evented_inotify,
             INOTIFY,
             mio::Ready::readable(),
-            mio::PollOpt::edge(),
+            mio::PollOpt::level(),
```
0xpr03, line 111: might want to add a comment here, referencing the issue

0xpr03 (approved):

src/inotify.rs
jimblandy: Sure.

jimblandy:
> I believe it's safe to backport, yes. If you want, I'm happy to accept it :)

Filed as #269.

PR #270, opened 2020-11-08
chore(deps): update crossbeam-channel to 5.0

Plecra:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ exclude = [
 [dependencies]
 anymap = "0.12.1"
 bitflags = "1.0.4"
-crossbeam-channel = "0.4.0"
+crossbeam-channel = "0.5.0"
 filetime = "0.2.6"
 libc = "0.2.4"
 serde = { version = "1.0.89", features = ["derive"], optional = true }


JohnTitor:
I don't think we could accept this as it will break our MSRV.

0xpr03:
@Plecra do you have a requirement to use a newer crossbeam-channel ?

Plecra:
Ah, bummer. I can close the PR if this just isn't suitable for `notify`, but maybe a new version release is a good time to make an MSRV bump? I've seen a lot of crates moving to 1.36.

I just spotted this when reviewing [bevy](https://github.com/bevyengine/bevy)'s duplicate dependencies. I can't speak for the project, but this would work towards https://github.com/bevyengine/bevy/pull/667 (`notify` is the only external crate that's using crossbeam-channel).

0xpr03:
I think we can do this for 0.5, as its still in preview.

JohnTitor:
Then it's fine to me as well. But we should tweak CI to make it green, and mention MSRV on the changelog since we'd like to treat it as a breaking change.

cart:
Let me know if theres anything I can do to help with getting this upgraded / releasing the next notify version. I'm invested in cleaning up Bevy's dep tree :smile:

PR #278, opened 2021-01-22
Update to mio 0.7, remove mio-extras

### LINKED ISSUE #248: upgrade to mio 0.7
New version of mio is out. Old version relies on a deprecated crate (net2).

roblabla:
This MR updates mio to version 0.7 and removes mio-extras.

To replace the channel used by mio-extras, we use crossbeam_channel along with a mio::Waker that will be used to wake up the mio event loop whenever we send a message through the crossbeam channel.

Note that  mio 0.7 has no level triggering (see https://github.com/tokio-rs/mio/issues/928 ). As such, we need another solution for #267. What I implemented is simply looping over `inotify.read_buffer` until it returns an empty iterator, at which point we can safely go back to waiting.

Fixes #248

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,8 +32,7 @@ walkdir = "2.0.1"
 
 [target.'cfg(target_os="linux")'.dependencies]
 inotify = { version = "0.8", default-features = false }
-mio = "0.6.15"
-mio-extras = "2.0.5"
+mio = { version = "0.7.7", features = ["os-ext"] }
 
 [target.'cfg(target_os="macos")'.dependencies]
 fsevent = "2.0.1"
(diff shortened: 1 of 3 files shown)


0xpr03:
Hey, thanks for the research and PR. Sadly we will need a solution for #267 first, otherwise this is a regression..

roblabla:
@0xpr03 this PR fixes #267 by repeatedly calling `INotify::read_buffer`, see https://github.com/notify-rs/notify/pull/278/commits/5c48373a5d72f4f5fe09e892247d1d0fcee4d29b

0xpr03 (approved):
As folks at tokio also approve the workaround I see no issues anymore with merging.

JohnTitor (approved):
Awesome, thanks for migrating! Looks great to me.

Issue #275, opened 2020-12-26
Docs example for hot reloading


vladinator1000:
Hi y'all, I've been trying to figure out how to reload a struct on a file change and was thinking the docs could really use an example of that. I got somewhere but my current solution panics [here](https://github.com/savovsa/announcer/commit/c21c0c0f36f4982f1092398094094720481ca05c#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR24) when you save the file too often. It seems that it's passing [nothing to Serde](https://github.com/savovsa/announcer/commit/c21c0c0f36f4982f1092398094094720481ca05c#diff-121f57fd9da9016a85108d2a879fdc4e8d03e3699857e7528d90f2199229153fR35) for deserialization.

Here's the error
`EOF while parsing a value at line 1 column 0`

I should also mention this is on Windows and it seems that windows fires 2 Modify Any events whenever you save the file.

0xpr03:
Depending on how your editor works its quite possible that your config file is truncated (empty) and then the new content is added. So for a brief moment an EOF is read as the file is empty.

vladinator1000:
@0xpr03 thank you, I added some error checks for that! Do you mind if I add a more complex example to the examples folder that looks a bit [like this](https://github.com/savovsa/announcer/blob/mutex/src/main.rs)?

0xpr03:
Hm, that may be ok. On the other side I'd hope that people are not using notify-rs for async stuff ? I mean tokio and async-std come with their own file watchers AFAIK?

vladinator1000:
Hmm I don't see any support for that in async-std. Notify-rs seems to work great for my use case of hot reloading so far.

0xpr03:
Then I'd be fine with adding an example. At best add some kind of comment to document what you're trying to accomplish (sync notify -> async code etc)

0xpr03:
Feel free to close this due t #279 being merged.

PR #279, opened 2021-01-24
Add a hot reloading example.

savovsa:
# A hot reloading web server

Imagine this is a web app that remembers information about audio messages.
It has a configuration file that acts as a database, you can edit the configuration and the app will pick up changes without the need to restart it.

You can use this concept to make your hot-reloading web servers. It uses a mutex under an atomic reference counted pointer to guarantee that the config won't be read and written to at the same time. 

The work on this example is in collaboration with @vladinator1000 in relation to https://github.com/notify-rs/notify/issues/275


0xpr03 (changes requested):
So far the code has nearly zero comments, which is kinda bad for people that are trying to figure out rust. Can you add some more info which file is doing what inside of them?

~~Could you add some readme.md that explains the purpose of this example ? (What does this application do, what files are doing what).~~ Github stop hiding stuff..

savovsa:
@0xpr03 We added some comments, do they look ok?

0xpr03 (approved):
That is far more in depth than I wanted but even better. I would've settled with "here we start a webserver, here is the part that hot-reloads the config" or something like this ;)

PR #281, opened 2021-02-14
Handle interrupted error from mio

rschoon:
mio 0.7 was changed to no longer automatically retry on system call interruption, so we now must do it ourselves.  For example, the process receiving SIGINT could cause notify to panic, if the signal handler had been changed from default:
```
thread '<unnamed>' panicked at 'poll failed: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }', /home/rschoon/.cargo/registry/src/github.com-1ecc6299db9ec823/notify-5.0.0-pre.5/src/inotify.rs:134:47
```

This change basically just ignores it (since we'll retry when we go through the loop again).

### DIFF
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -131,7 +131,14 @@ impl EventLoop {
         let mut events = mio::Events::with_capacity(16);
         loop {
             // Wait for something to happen.
-            self.poll.poll(&mut events, None).expect("poll failed");
+            match self.poll.poll(&mut events, None) {
+                Err(ref e) if matches!(e.kind(), std::io::ErrorKind::Interrupted) => {
+                    // System call was interrupted, we will retry
+                    // TODO: Not covered by tests (to reproduce likely need to setup signal handlers)
+                },
+                Err(e) => panic!("poll failed: {}", e),
+                Ok(()) => {}
+            }
 
             // Process whatever happened.
             for event in &events {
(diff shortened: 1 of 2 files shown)


0xpr03 (commented):
Could we test this somehow in CI ?

rschoon:
Probably, although getting it right seems like it might be a bit complicated.  Doing it as an integration test and sending a signal to its own pid would probably be sufficient to exercise the problem, but verifying things went according to plan might be harder.

src/inotify.rs
```diff
@@ -131,7 +131,13 @@ impl EventLoop {
         let mut events = mio::Events::with_capacity(16);
         loop {
             // Wait for something to happen.
-            self.poll.poll(&mut events, None).expect("poll failed");
+            match self.poll.poll(&mut events, None) {
+                Err(ref e) if matches!(e.kind(), std::io::ErrorKind::Interrupted) => {
+                    // System call was interrupted, we will retry
```
0xpr03, line 136: I'd like to land this but not loose the information, could you just document in the comment that there is no actual test for this ?

0xpr03 (changes requested):

0xpr03 (approved):

JohnTitor (approved):
LGTM, I think it's worth mentioning this in the changelog, could you add an entry?

@0xpr03 Feel free to merge once it's done :)

0xpr03:
Thanks for tackling this !

PR #284, opened 2021-02-20
Prepare for 5.0.0-pre.6

0xpr03:
This also updates the lib version in the docs, which we forgot to update since pre.2 apparently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
 
 ## unreleased
 
+
+
+## 5.0.0-pre.6 (2021-02-20)
+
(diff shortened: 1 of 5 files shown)


JohnTitor (approved):
> This also updates the lib version in the docs, which we forgot to update since pre.2 apparently.

👍, could you also update README, like https://github.com/notify-rs/notify/commit/f48257f0343aa56defba003bfdcd89633b999666? Otherwise LGTM.

0xpr03:
Ah yeah. I just added a release_checklist.md so this hopefully will be smoother in the future.

PR #285, opened 2021-03-11
Display correct error on max-files-watch

### LINKED ISSUE #266: 'Error: No space left on device (os error 28)' when watching too many files
#### System details

- OS/Platform name and version: `Linux build-server 4.19.0-10-amd64 #1 SMP Debian 4.19.132-1 (2020-07-24) x86_64 GNU/Linux`
- Rust version (if building from source): `rustc --version`: `rustc 1.49.0-nightly (ffe52882e 2020-10-30)`
- Notify version (or commit hash if building from git): 5.0.0-pre.3

- If you're coming from a project that makes use of Notify, what it is, and a link to the downstream issue if there is one:  See https://github.com/Canop/bacon/issues/10.
- Filesystem type and options: ext4, errors=remount-ro
- If you're running as a privileged user (root, System): no

I ran `notify` on `src/` of rust-lang/rust with `RecursiveMode::Recursive`.

#### What you expected

Either it should work, or `notify` should give a helpful error when trying to watch too many files.

#### What happened

`notify` printed `Error: No space left on device (os error 28)`.

0xpr03:
This is a draft to display the correct error message when maximum number of files watchable is reached. Under linux you can test this by using the default example and setting `sudo sysctl -w fs.inotify.max_user_watches=1`

It currently works as expected on
- [X] linux
- [x] Windows
- [x] MacOS

Windows does not seem to expose some kinds of limits on its API.

I don't have access to a mac system ~~but they have [limit](https://superuser.com/questions/630777/how-do-i-configure-fsnotifys-queue-size-watch-limit-on-osx), so it'd be nice to know how its working over there (possibly the same due to libc?). Windows will probably behave completely different.~~ See down below for why macos doesn't seem to need any changes.

Fixes #266

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -29,6 +29,9 @@ pub enum ErrorKind {
 
     /// An invalid value was passed as runtime configuration.
     InvalidConfig(Config),
+
+    /// Can't watch (more) files, limit on the total number of inotify watches reached
+    MaxFilesWatch,
 }
 
 /// Notify error type.
@@ -75,6 +78,11 @@ impl Error {
 
     /// Creates a new i/o Error from a stdlib `io::Error`.
     pub fn io(err: io::Error) -> Self {
+        // do not report inotify limits as "no more space" on linux #266
+        #[cfg(target_os = "linux")]
+        if err.raw_os_error() == Some(28) {
+            return Self::new(ErrorKind::MaxFilesWatch);
+        }
         Self::new(ErrorKind::Io(err))
     }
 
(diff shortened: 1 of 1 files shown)


0xpr03:
On a separate note: We might want to have a list of people than can test stuff like [glutin/rust-windowing](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors) has.

0xpr03:
This is blocked by a mac implementation (possibly just another cfg with its error no) which I can't do as I don't have any apple machine.

JohnTitor (approved):
Looks reasonable to me, though I also don't have a macOS machine :/ Could we check it with a (temporary) test on CI?

0xpr03:
With the help of a friend I've come to the following conclusion: There doesn't seem to be an watch-files limit on mac. And if there would be, you would not get any userland reports.. Also the only real limits seems to be how many watcher you can spawn in parallel. Apparently 2047 is the limit, though this could also be a thread limit.

Thus I'd only add the linux change and leave everything else as is.

PR #286, opened 2021-03-14
Add debouncer

0xpr03:
Add debouncer back to v5.


0xpr03:
Open questions:

- Do we want to add multiple-receiver functionality?
- What is the priority for any/other events ?
- Do we allow much configuration in light of "Any" on invalid configuration overriding anything?
- Do we rather unify the debouncer and configuration into one thing ?
- Do we want to spawn another thread for timing? Otherwise we'll need more redesign.

0xpr03:
Another I just realized: We need an idea how we handle configuration changes for debouncer and immediate watchers. Currently we don't care at all. Historically every debouncer init was implemented at the specific notify modules (inotify, fsevent..).

What I realized: we don't actually implement any of the current config options. Or am I wrong ? Searching for PreciseEvents doesn't emit one hit across any of the specific implementations. Also from all the config options exactly one is relevant to the specific implementations of the immediate watchers: PreciseEvents, everything else is related to the debouncer. Thus the whole ".config()" operation is nearly irrelevant for immediate watchers.

cc @JohnTitor

IWANABETHATGUY:
I am curious when will this merge into master?

0xpr03:
Once I've got time to fully undertake this as there is a lot I want to get right or at least decide actively to not include it, as you can read above. Which due to private things going on could take until march next year. I consider v4 to be pretty stable, so feel free to use that one.

0xpr03:
This is the tiniest debouncer I could imagine being useful. Only Any events are delivered and AnyContinuous for ongoing events past the timeout.

Things I don't like:
- you can't change the tick rate
- you can't change the timeout

0xpr03:
The real problem with all of these three is how we pass the configuration stuff back and forth. User -> backend (config), User -> debouncer (config), backend -> debouncer (events, errors), debouncer -> User (events, errors)

src/debouncer.rs
```diff
@@ -0,0 +1,198 @@
...
+/// Timeout is the amount of time after which a debounced event is emitted or a Continuous event is send, if there still are events incoming for the specific path.
+/// 
+/// If tick_rate is None, notify will select a tick rate that is less than the provided timeout.
+pub fn new_debouncer(
+    timeout: Duration,
+    tick_rate: Option<Duration>,
+) -> Result<(Receiver<DebounceChannelType>, RecommendedWatcher), Error> {
+    let data = DebounceData::default();
+
+    let tick_div = 4;
```
JohnTitor, line 132: nit: I'd like to manage this value as const and declare it on the module scope.

JohnTitor (approved):
I'm not too familiar with it but the implementation looks great to me with a nit!

src/debouncer.rs
0xpr03: understandable, had a moment of wrangling with myself whether I'd want that as top module declaration

visig9:
I believe configurations should not be given in general level. Ref: https://docs.rs/notify/5.0.0-pre.15/notify/enum.Config.html

- `NoticeEvents`: Applicable to debounced watchers only.
- `OngoingEvents`: Applicable to debounced watchers only.
- `PreciseEvents`: The exactly meaning depend on watcher implemention detail. That mean user must know which watcher they want to used.

Further more, current interface also force API user need to understand which config option should be set in which location. For example if I want to use `PollWatcher`, I need to check following configuration location:

- Set `PreciseEvents` here: https://docs.rs/notify/5.0.0-pre.15/notify/trait.Watcher.html#method.configure
- Set `poll_interval` & `compare_contents` in here: https://docs.rs/notify/5.0.0-pre.15/notify/poll/struct.PollWatcher.html#method.with_config
- Set `RecursiveMode` in here: https://docs.rs/notify/5.0.0-pre.15/notify/trait.Watcher.html#tymethod.watch
    - I know this is watch point level config but feel like a config and not easy to set. Suggestion: how about just split `.watch(path, recu_mode)` into 2 function like `.watch_exact(path)`, `.watch_recu(path)`.

If do so, `notify` can give every single watcher implementation a dedicated builder to let API user easy to config everything what they want.

IWANABETHATGUY:
Any progress for this pull request? Could we merge this?

CrendKing:
Why is the `Receiver` hardcoded to `std::sync::mpsc` again? I thought the v5 was meant to decouple the channel from the watcher.

0xpr03:
This got a little bit bigger now. Fixing the audit issues and decoupling the debouncer from notify core.

0xpr03:
As good as finished.
Open points:
- how can I enable custom watcher options (see pollwatcher) with this system ? Currently I can't think of a good combination.

passcod (approved):

PR #288, opened 2021-03-18
V4 leak fix

### LINKED ISSUE #287: Window's notify thread is stuck hanging
#### System details

- OS/Platform name and version: Windows 10
- Rust version (if building from source): `rustc --version`: rustc 1.50.0
- Notify version (or commit hash if building from git): 4.0.15

We created a bunch of `ReadDirectoryChangesWatcher` thru `notify::watcher`. After we dropped the `ReadDirectoryChangesWatcher`s, we expected the watcher to drop the given `mspc::Sender`.

#### What you expected

Sometimes, the given `mspc::Sender` wouldn't be dropped by the `ReadDirectoryChangesWatcher` after the `ReadDirectoryChangesWatcher` was dropped.

#### What happened

I tracked it down to a race condition in the notify crate (4.0.15) under [src/debounce/timer.rs](https://github.com/notify-rs/notify/blob/4c880c6a7259b54c4a838b2494199827c2151cd7/src/debounce/timer.rs).
A particular interleaving of operations could leave the `ScheduleWorker` thread alive while the `WatchTimer` thread has dropped.

Below is a simplified model of WatchTimer and ScheduleTimer showing the ScheduleTimer hanging after WatchTimer drops:
```rust
use std::sync::{atomic::AtomicBool, Arc, Condvar, Mutex};

fn main() {
    // We will use a channel to enforce an interleaving that shows the thread hang
    let (rendezvous_sender, rendezvous_receiver) = std::sync::mpsc::sync_channel::<()>(0);

    // Below is a modeling of WatchTimer::new spawning ScheduleWorker
    let condvar = Arc::new(Condvar::new());
    let stopped = Arc::new(AtomicBool::new(false));
    let thread = spawn_schedule_worker(condvar.clone(), stopped.clone(), rendezvous_receiver);

    // Interleaving sync point [1]
    // Makes it so we pass the `stopped` AtomicBool check in the thread
    rendezvous_sender.send(()).unwrap();

    // Below is a modeling of WatchTimer::drop
    stopped.store(true, std::sync::atomic::Ordering::SeqCst);
    condvar.notify_one();

    // Interleaving sync point [2]
    // Makes it so we sent a notify on the condvar before the thread waits on it
    rendezvous_sender.send(()).unwrap();

    // Ensure the thread is terminated
    thread.join().unwrap();
}

fn spawn_schedule_worker(
    condvar: Arc<Condvar>,
    stopped: Arc<AtomicBool>,
    rendezvous_receiver: std::sync::mpsc::Receiver<()>,
) -> std::thread::JoinHandle<()> {
    std::thread::spawn(move || {
        let m = Mutex::new(());

        let mut g = m.lock().unwrap();

        loop {
            if stopped.load(std::sync::atomic::Ordering::SeqCst) {
                println!("Leaving loop via stopped bool");
                break;
            }

            // Interleaving sync point [1]
            // Makes it so we pass the `stopped` AtomicBool check in the thread
            rendezvous_receiver.recv().unwrap();

            // Interleaving sync point [2]
            // Makes it so we sent a notify on the condvar before the thread waits on it
            rendezvous_receiver.recv().unwrap();

            println!("Waiting for condvar");
            g = condvar.wait(g).unwrap();
            println!("Waking from condvar");
        }
    })
}
```

The creation of `condvar`, `stopped`, and `thread` are analogous to [`WatchTimer::new`](https://github.com/notify-rs/notify/blob/4c880c6a7259b54c4a838b2494199827c2151cd7/src/debounce/timer.rs#L155) and storing true in `stopped` and using `condvar.notify_one()` is analogous to [`WatchTimer::drop`](https://github.com/notify-rs/notify/blob/4c880c6a7259b54c4a838b2494199827c2151cd7/src/debounce/timer.rs#L201).

A thread interleaving is enforced via a rendezvous channel. Running this above rust program should show a program hang after "Waiting for condvar" is printed.

Awfa:
Fixes #287


src/windows.rs
```diff
@@ -235,7 +235,9 @@ fn stop_watch(ws: &WatchState, meta_tx: &Sender<MetaEvent>) {
         let ch = handleapi::CloseHandle(ws.dir_handle);
         // have to wait for it, otherwise we leak the memory allocated for there read request
         if cio != 0 && ch != 0 {
-            synchapi::WaitForSingleObjectEx(ws.complete_sem, INFINITE, TRUE);
+            while synchapi::WaitForSingleObjectEx(ws.complete_sem, INFINITE, TRUE) != WAIT_OBJECT_0
+            {
```
0xpr03, line 239: If I'm correct this cold be a hot loop ? Because I'd rather sleep than let this run hot.

0xpr03 (changes requested):
Thanks for the work!

0xpr03:
Note: I'll do a v4 release once this is merged.

src/windows.rs
Awfa: I don't think it's a hot loop.

This loop will cause the thread to either handle the APC queue (in which case it will return WAIT_IO_COMPLETION), or once the APC which got queued from the CloseHandles above gets handled, will finish by meeting the WAIT_OBJECT_0 condition.

Sleeping will be unproductive since for the complete_sem to be in a signaled state, the APC that will ReleaseSemaphore will need to be called. And the only way to call and make progress on that front - is to WaitForSingleObjectEx to consume the APC queue.

0xpr03 (approved):

JohnTitor (commented):
Could we add some tests somehow? I don't think v4 is actively developed but I'd like to avoid the regression by a further patch.

Awfa:
> Could we add some tests somehow? I don't think v4 is actively developed but I'd like to avoid the regression by a further patch.

Hmm - I don't know how we can trigger the specific race using the public interface of `ReadDirectoryChangesWatcher` - which the test cases use.

Awfa:
@0xpr03 do you know if those test failures are from just flaky tests? I remember the previous test run didn't have those failures.

0xpr03:
You can ignore macos tests.

JohnTitor:
> Hmm - I don't know how we can trigger the specific race using the public interface of `ReadDirectoryChangesWatcher` - which the test cases use.

Hm fair enough, anyway the fix is obvious so I'm fine as-is.

> do you know if those test failures are from just flaky tests? 

Yeah, currently GHA doesn't support the allow-failure feature.

JohnTitor (approved):

Awfa:
Alright - what's the next step to get this PR in? I'm not sure how to rerun the checks.

JohnTitor:
@0xpr03 Feel free to merge this (and make a new release if needed) if you think we don't need further action here.

0xpr03:
Yeah I'll have to rebase this later to merge it, according to github.

PR #277, opened 2021-01-06
Setup GitHub Actions

JohnTitor:
Closes #271

PR #290, opened 2021-03-19
Backport #285 display correct inotify limit error

0xpr03:
Backport #285 to v4.


CHANGELOG.md
```diff
@@ -85,7 +87,7 @@
 - META: Change commit message style: commits are now prefixed by a `[topic]`.
 - FIX: Make sure debounced watcher terminates. [#170]
 - FIX: \[Linux\] Remove thread wake-up on timeout (introduced in 4.0.5 by error). [#174]
-- FIX: Restore compatibility with Rust before 1.30.0. [`eab75118`] 
+- FIX: Restore compatibility with Rust before 1.30.0. [`eab75118`]
```
0xpr03, line 90: my editor doesn't seem to like this line..

JohnTitor (approved):
👍 

> I think it'd be better to introduce a new error case, but this approach technically does not break semver.

We're going to v5 so this is still reasonable for me as a "patch".

0xpr03:
> 
> 
> 👍
> 
> > I think it'd be better to introduce a new error case, but this approach technically does not break semver.
> 
> We're going to v5 so this is still reasonable for me as a "patch".

I'm not sure I understand what you mean. That we're going towards a stable v5 ? Because this is the backport for v4.

src/lib.rs
```diff
@@ -520,6 +520,11 @@ impl StdError for Error {
 
 impl From<io::Error> for Error {
     fn from(err: io::Error) -> Error {
+        // do not report inotify limits as "no more space" on linux #266
+        #[cfg(target_os = "linux")]
+        if err.raw_os_error() == Some(28) {
+            return Error::Generic(String::from("Can't watch (more) files, limit on the total number of inotify watches reached"))
+        }
```
JohnTitor, line 527: Btw, I noticed this breaks our MSRV.

```
error: attributes are not yet allowed on `if` expressions
   --> src/lib.rs:524:9
    |
524 |         #[cfg(target_os = "linux")]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
error: aborting due to previous error
```

So it should be:

```suggestion
        #[cfg(target_os = "linux")]
        {
        	if err.raw_os_error() == Some(28) {
            	return Error::Generic(String::from("Can't watch (more) files, limit on the total number of inotify watches reached"))
            }
        }
```

JohnTitor:
I mean, we don't _have to_ introduce a new error structure for v4 as it's passive-developed now (sorry if my wording is strange, I'm not a native speaker :/).

JohnTitor (changes requested):

0xpr03:
Ah ok makes sense, I'm not a native one either ;) I'll fix that suggestion and then merge.

JohnTitor:
Ugh:
image

I forgot to backport #277 to v4, will do later.

PR #297, opened 2021-04-14
Release v4.0.16

0xpr03:
Release 4.0.16 with all the changes, also finally update lib.rs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
-## 4.0.16 (future)
+## 4.0.17 (future)
+
+
+## 4.0.16 (2021-04-14)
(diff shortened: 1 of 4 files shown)


0xpr03:
I think we should stop using exact version pinning and at least pin only the minor version as we follow semver.

```toml
notify = "4.0"
```

JohnTitor:
> 
> 
> I think we should stop using exact version pinning and at least pin only the minor version as we follow semver. (And everything critical _should_ use lock files.)
> 
> ```toml
> notify = "4.0"
> ```

I don't have the strong opinion here but it should reduce the version bump work 👍

JohnTitor (approved):

PR #298, opened 2021-04-14
Fix leak on windows, porting #288 to v5

0xpr03:
This should fix the windows specific part of #287

cc @Awfa


Awfa:
Looks correct to me.

JohnTitor (approved):
LGTM!

PR #302, opened 2021-05-03
Check for errors from `str_path_to_cfstring_ref`.

### LINKED ISSUE #301: "fatal runtime error: Rust cannot catch foreign exceptions" with `remove_dir` on Mac
#### System details

- OS/Platform name and version: macOS Catalina (`Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64`)
- Rust version (if building from source): `rustc --version`: `rustc 1.51.0 (2fd73fabe 2021-03-23)`
- Notify version (or commit hash if building from git): Current `main` (4a3f5a35d15dbed5611732016f90c785fdf0f0ac). I observed the same behavior using the debounced `notify::watcher` API in Notify 4.0.16.
- I'm not running as a privileged user, or in a container, or in a VM.
- Filesystem type and options: Here's the output of `mount`:

    ```
    /dev/disk1s1 on / (apfs, local, read-only, journaled)
    devfs on /dev (devfs, local, nobrowse)
    /dev/disk1s2 on /System/Volumes/Data (apfs, local, journaled, nobrowse)
    /dev/disk1s5 on /private/var/vm (apfs, local, journaled, nobrowse)
    map auto_home on /System/Volumes/Data/home (autofs, automounted, nobrowse)
    /dev/disk1s4 on /Volumes/Recovery (apfs, local, journaled, nobrowse)
    ```

    On this system, the directory that is actually watched is `"/var/folders/vr/my8x9tmn4y7cdjcj8f7jw2gr0000gn/T/.tmprPcUcB"`, and `/var` is a symlink to `private/var`.

1. clone the `notify` repo
2. make a file `tests/race-with-remove-dir.rs` with the contents below
3. `cargo test`

```rust
use std::{env, fs, thread, time::Duration};

use notify::{immediate_watcher, RecursiveMode, Watcher};

#[test]
fn test_race_with_remove_dir() {
    let tmpdir = env::temp_dir().join(".tmprPcUcB");
    fs::create_dir_all(&tmpdir).unwrap();

    {
        let tmpdir = tmpdir.clone();
        thread::spawn(move || {
            let mut watcher = immediate_watcher(move |result| {
                eprintln!("received event: {:?}", result);
            })
            .unwrap();

            watcher.watch(tmpdir, RecursiveMode::NonRecursive).unwrap();
        });
    }

    let subdir = tmpdir.join("146d921d.tmp");
    fs::create_dir_all(&subdir).unwrap();
    fs::remove_dir_all(&tmpdir).unwrap();
    thread::sleep(Duration::from_secs(1));
}
```

#### What you expected

The test might pass, fail, or panic. I'm fine with any of those outcomes.

#### What happened

```console
     Running target/debug/deps/race_with_remove_dir-2738b612224032ad

running 1 test
fatal runtime error: Rust cannot catch foreign exceptions
error: test failed, to rerun pass '--test race-with-remove-dir'

Caused by:
  process didn't exit successfully: `/Users/jorendorff/repos/notify/target/debug/deps/race_with_remove_dir-2738b612224032ad` (signal: 6, SIGABRT: process abort signal)
```

jorendorff:
Closes #301.

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -281,6 +281,10 @@ impl FsEventWatcher {
         unsafe {
             let mut err: cf::CFErrorRef = ptr::null_mut();
             let cf_path = cf::str_path_to_cfstring_ref(str_path, &mut err);
+            if cf_path.is_null() {
+                cf::CFRelease(err as cf::CFRef);
+                return Err(Error::watch_not_found().add_path(path.as_ref().into()));
+            }
 
             let mut to_remove = Vec::new();
             for idx in 0..cf::CFArrayGetCount(self.paths) {
@@ -320,6 +324,12 @@ impl FsEventWatcher {
         unsafe {
             let mut err: cf::CFErrorRef = ptr::null_mut();
             let cf_path = cf::str_path_to_cfstring_ref(str_path, &mut err);
+            if cf_path.is_null() {
+                // Most likely the directory was deleted, or permissions changed,
+                // while the above code was running.
+                cf::CFRelease(err as cf::CFRef);
+                return Err(Error::path_not_found().add_path(path.as_ref().into()));
+            }
             cf::CFArrayAppendValue(self.paths, cf_path);
             cf::CFRelease(cf_path);
         }
--- a/tests/race-with-remove-dir.rs
+++ b/tests/race-with-remove-dir.rs
@@ -1,0 +1,28 @@
+use std::{env, fs, thread, time::Duration};
+
+use notify::{immediate_watcher, RecursiveMode, Watcher};
+
+/// Test for <https://github.com/notify-rs/notify/issues/301>.
+/// Note: This test will fail if your temp directory is not writable.
+#[test]
+fn test_race_with_remove_dir() {
+    let tmpdir = env::temp_dir().join(".tmprPcUcB");
+    fs::create_dir_all(&tmpdir).unwrap();
+
+    {
+        let tmpdir = tmpdir.clone();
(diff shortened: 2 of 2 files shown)


src/fsevent.rs
```diff
@@ -281,6 +281,10 @@ impl FsEventWatcher {
         unsafe {
             let mut err: cf::CFErrorRef = ptr::null_mut();
```
0xpr03, line 282: we should probably investigate whether this should be handled instead of ignored (and deallocated)

0xpr03 (approved):
Are you sure this fixes it or rather just works in case the race condition for our runtime panic isn't triggered ?

jorendorff:
> Are you sure this fixes it or rather just works in case the race condition for our runtime panic isn't triggered ?

Medium-high confidence that this fixes the actual problem. Here's the evidence:

- Before the change, the test crashed pretty reliably for me, say 95-98% of test runs, except in `lldb`. The first thing I tried is just adding `assert!(!cf_path.is_null())` there, and it stopped the crashes immediately; they became regular Rust panics.

- While I was reducing this from my real application, I ended up making 8 versions of the test (each containing fewer and fewer random bits of my app) and the change, backported to 4.0, makes them all pass.

To get a bit more confidence, one could write a program that passes a null second argument to `cf::CFArrayAppendValue`, to see if it crashes with this strange message about a "foreign exception".

PR #303, opened 2021-05-04
Backport #302 to v4

0xpr03:
(cherry picked from commit 86421e513ee8b5b315648faec5da0ed5eb114d39)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 4.0.17 (future)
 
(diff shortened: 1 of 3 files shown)


0xpr03:
@JohnTitor any ideas why windows CI setup fails currently ? same on main branch

JohnTitor:
> any ideas why windows CI setup fails currently ? same on main branch

Posted #304 and #305 to fix that issue. So, rustup now tries to self-update but it overrides the rustup binary, then causes a permission failure.

PR #304, opened 2021-05-06
Do not self-update rustup to avoid permission failure

JohnTitor:
The current Rustup version on GHA is 1.23.1 so rustup tries to self-update when installing a toolchain but it can cause a permission issue on Windows env. `--no-self-update` is needed to avoid it.


0xpr03:
thanks!

PR #307, opened 2021-05-10
Remove anymap

### LINKED ISSUE #306: Consider migrating off anymap, it contains undefined behavior
It appears that the `anymap` crate contains some [undefined behavior](https://github.com/chris-morgan/anymap/pull/32) that has been fixed in the repository, but was never published to crates.io. Unfortunately it appears the project has since been abandoned (https://github.com/chris-morgan/anymap/issues/37, https://github.com/RustSec/advisory-db/pull/906).

I'd recommend migrating off the project, especially since there still may be an opportunity to make api breaking changes before 5.0.0 is released. Unfortunately there isn't an officially recommended solution. There's https://github.com/azriel91/anymap2, which is a fork that includes the fix, but I'd reconsider a more straightforward approach, where the event type uses [serde-value](https://docs.rs/serde-value) to communicate arbitrary values.

```rust
struct Event {
    ...

    // Use an option-box to make it a single pointer sized
    pub attrs: Option<Box<EventAttrs>>,
}

struct EventAttrs {
    pub tracker: Tracker,
    pub info: Info,
    pub flag: Flag,
    pub extra: HashMap<String, serde_value::Value>,
}
```

Or if you want to still support arbitrary types, you could implement your own AnyMap, since TypeId is hashable.

erickt:
`anymap` has some undefined behavior, and appears to no longer be maintained.
This patch replaces anymap with an opaque `Attributes` type, which contains a
number of accessor methods to get and set values. Under the covers, this type
is simply a box of a struct of option types. However, since it's opaque the
underlying type could be swapped out with an alternative type that can be
extended to support more attributes without breaking users.

Closes #306


0xpr03 (changes requested):
Thanks for the work. Can you add a changelog entry ? See [this](https://github.com/notify-rs/notify/commit/3ea4e8f350947c92815bdaf52a67c1d52131ca6f) for example

0xpr03 (approved):

PR #312, opened 2021-05-12
Address clippy lints, hide some fsevents functions

erickt:
This addresses all the items found by `clippy 0.1.54 (ca82264e 2021-05-09)`. In addition, it hides the functions `fsevents::Callback` and `fsevents::CFRunLoopIsWaiting`, which I don't think were intended to be made public.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,11 @@
 - CHANGE: Avoid stating the watched path for non-recursive watches with inotify [#256]
 - FIX: Report events promptly on Linux, even when many occur in rapid succession. [#268]
 - FIX: Remove `anymap`, and replace event attributes with an opaque type. [#306]
+- CHANGE: Hide fsevent::{CFRunLoopIsWaiting,callback}, fix clippy lint warnings [#312]
 
 [#268]: https://github.com/notify-rs/notify/pull/268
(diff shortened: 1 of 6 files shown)


0xpr03:
I'm not sure why the macos-latest run fails only for this PR with a timeout after 46minutes. I retried the CI and it failed again.

erickt:
I had a bad cast in the callback, which I've fixed. I've added two more fix:

* I found and fixed a memory leak (found with `RUSTFLAGS=-Zsanitizer=leak cargo test --target x86_64-apple-darwin)`.
* I rewrote how CFRefs were being passed between threads in order to avoid casts, and moved freeing the context into the thread after the thread was shut down to avoid the risk that we could add a path that freed the context out from under the event stream.

0xpr03:
Thanks for all the work, this looks pretty well. I think we are really in need of somebody that has a mac. Do you want this merged now or are you planning to add more ?

Also please also add a changelog entry like previously for the "unreleased" section with a link to your PR. Or just leave a note and I'll do that.

0xpr03:
Also I'll have to check, the public stuff could be from the deleted tests we've had previously.

erickt:
@0xpr03 - thanks! added a CHANGELOG comment. I think this PR is ready to go. I'll rebase #313 on top of this once it lands.

PR #313, opened 2021-05-12
Remove dependency on fsevent

erickt:
Notify doesn't really use the `fsevent` crate, so this patch removes the dependency. It still uses `fsevent-sys` to provide the C wrappers for the fsevent library.


erickt:
I spoke too soon! @octplane merged in some of my PRs, so I’ll refactor this branch on top of that instead.

erickt:
I've updated this PR to just remove the `fsevent` crate (which isn't really used in `notify`).

PR #317, opened 2021-05-12
Prepare 5.0.0-pre.8

0xpr03:
Release pre.8 to hotfix current dependency problem


JohnTitor (approved):
👍, thanks for taking care of it, @erickt and @0xpr03!

PR #319, opened 2021-05-13
Change CFRunLoopIsWaiting to return cf::Boolean

erickt:
According to this [doc], it's undefined behavior for a `bool` to be any
value other than `0` or `1`. Since `Boolean` is ultimately an
`unsigned char`, this avoids the risk that this function could return a
`true` value that's non-zero but not `1`.

[doc]: https://doc.rust-lang.org/reference/behavior-considered-undefined.html


0xpr03:
Thanks!

0xpr03:
@erickt as you did all of the work, would you prefer a release  for these last 3 PRs now ? Otherwise I'd wait until we gathered some more changes.

Issue #311, opened 2021-05-11
Future proofing 5.0.0 EventKind with #[non_exhaustive]


erickt:
More follow on to #306, what do you think about marking all the `EventKind` enum types, and child enums as `#[non_exhaustive]`? Looking at both the [fsevents flags](https://developer.apple.com/documentation/coreservices/1455361-fseventstreameventflags) and [inotify manpage](https://man7.org/linux/man-pages/man7/inotify.7.html), it looks like the event type expands over time, so I think it'd be worth allowing new event types to be added over time.

I think this would be preferable over using `Event::set_info`, at least in most cases, since it can be difficult for downstream users to understand the stability guarantee of the info strings, especially since they are not documented.

0xpr03:
Apparently I didn't send my thoughts on this..

I think it's ok to mark it as non_exuastive, though I'm not sure what the exact consequences for users is. I'm not sure if you could call it counter position, but in other discussions [people mentioned](https://github.com/notify-rs/notify/issues/187#issuecomment-477874045) that we would probably steer towards anonymous events. Meaning that there are users that don't care about the exact type and just need the knowledge that anything happened.

0xpr03:
(The current API in general is, depending on which issue you read, completely wrong or completely what the people want..)

erickt:
I don't have firm opinions on this either (especially since I haven't actually had to use this library for real yet :) ). The main consequence would be that users would have to write their event type handlers as:

```
match event.kind {
    EventKind::Any => { ... },
    EventKind::Access(_) => { ... },
    EventKind::Create(_) => { ... },
    EventKind::Modify(_) => { ... },
    EventKind::Remove(_) => { ... },
    EventKind::Other => { ... }
    // always need to handle the case that we got some event that we don't know about.
    _ => { ... }
}
```

And likewise for the other sub-kinds.

The other strategy would be to declare constants for the fsevents info() (and other future users of info()).  It's not perfect, but it helps to ameliorate the risk that these info strings change out from under a user. Or going even further, we could change the `Other` field to be parameterized by a `Box<dyn Any>`, which would be some watcher-specific field, but then we need to pay for the box for non-standard things.

Like I said though, I don't have enough experience with notify, and all the underlying OS implementations, to have a great recommendation, so I'd be fine letting this percolate over time.

Issue #310, opened 2021-05-10
Consider hiding fields in Event


erickt:
As part of developing #307, I realized it might be worth hiding the public fields in `Event` as part of 5.0.0. This would allow the underlying object to be extended with new attributes without breaking existing clients. I have a proof-of-concept of this change in https://github.com/erickt/notify/tree/hide-fields that's built upon #307 that I can push up after we address #306.

0xpr03:
I've got no real stakes in this, we've had no problems with them being public so far. Also your current implementation adds an additional .clone - though its probably irrelevant. So I'm ok with merging it.

PR #320, opened 2021-05-17
Fix clippy lints

erickt:
This fixes a few lint violations that came from my recent set of changes.

With this, would it be possible to cut a 5.0.0-pre.9? I think my tickets #310 and #311 could percolate a little longer, and don't need to be addressed for my needs.

Thanks so much!


0xpr03 (changes requested):
Looks good, I'll cut a release for that. Can you add an changelog entry in the `## unreleased` section ?

erickt:
Oops forgot that again. Done!

0xpr03 (approved):

PR #321, opened 2021-05-18
Publish 5.0.0-pre.9

0xpr03:


erickt:
Could this be updated to incorporate #322? Then we won't need to release a new version of notify to incorporate updates to fsevent-sys.

0xpr03:
Sorry for raking so long, it's released

PR #325, opened 2021-06-02
Fix StreamContextInfo to be Send

### LINKED ISSUE #324: Soundness Bug - StreamContextInfo is not safe to send across threads
We found another soundness issue. In src/fsevent.rs we have this definition:

```
struct StreamContextInfo {
    event_fn: Arc<dyn EventFn>,
    recursive_info: HashMap<PathBuf, bool>,
}
```

This type is passed across a thread that actually listens to file changes. Unfortunately this type isn't actually send because while `pub trait EventFn: 'static + Fn(Result<Event>) + Send {}` implements `Send`, `Arc<dyn EventFn>` is not `Send`. That also requires `EventFn` to also implement `Sync`. Unfortunately just adding `Sync` doesn't compile because of some other conflicts.

erickt:
In #324, we found that we were passing StreamContextInfo to another thread, even though the type is not actually Send. This makes sure that StreamContextInfo is actually Send by wrapping it in a mutex.

Fixes #324

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -457,7 +457,7 @@ impl FsEventWatcher {
                 .send(())
                 .expect("error while signal run loop is done");
         });
-        // block until runloop has been set
+        // block until runloop has been sent
         self.runloop = Some((rl_rx.recv().unwrap().0, done_rx));
 
         Ok(())
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -25,7 +25,7 @@ use std::ffi::CStr;
 use std::os::raw;
 use std::path::{Path, PathBuf};
 use std::ptr;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::thread;
 
 bitflags::bitflags! {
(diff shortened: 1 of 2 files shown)


src/fsevent.rs
```diff
@@ -531,14 +534,14 @@ unsafe fn callback_impl(
         for ev in translate_flags(flag, true).into_iter() {
             // TODO: precise
             let ev = ev.add_path(path.clone());
-            (*event_fn)(Ok(ev));
+            (event_fn.lock().unwrap())(Ok(ev));
```
0xpr03, line 537: I'd use an expect with a declaration what failed.

0xpr03 (changes requested):

src/fsevent.rs
erickt: Switched to `expect()`

0xpr03: thanks, I meant more something about event_fn having this poisoning, but maybe that's not good

0xpr03 (approved):

PR #330, opened 2021-06-04
Only check for ENOSPC on inotify_add_watch

erickt:
According to the inotify documentation, the only function that returns `ENOSPC` is `inotify_add_watch`, so this moves the check for this error code to the `inotify.add_watch()` call point. This makes sure someone using a different watcher like PollWatcher on linux won't reinterpret this error.

Also, this uses `libc::ENOSPC` instead of `28` because it's easier to interpret from the inotify documentation.


JohnTitor (approved):
Makes sense to me 👍

0xpr03 (approved):
thanks

PR #333, opened 2021-06-08
Change EventFn to take FnMut

erickt:
This changes EventFn to work with FnMut, rather than Fn. The advantage of this is that this allows callers to use state in the EventFn, rather than having to use interior mutability. This is helpful particularly for getting notify to work with async rust, since `futures::channel::mpsc`'s methods all take `&mut self`.

The downside of this change is that it limits the ability of notify to share `EventFn` across multiple threads, but all the backends either guarantee only a single background thread, or they already use an `Arc<Mutex<...>>` to protect `EventFn`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
 
 ## unreleased
 
+- CHANGE: Change EventFn to take FnMut [#333]
+
+[#333]: https://github.com/notify-rs/notify/pull/333
+
 ## 5.0.0-pre.10 (2021-06-04)
 
 - FIX: Make StreamContextInfo `Send` to fix soundness issue [#325]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ fsevent-sys = "4"
 winapi = { version = "0.3.8", features = ["fileapi", "handleapi", "ioapiset", "minwinbase", "synchapi", "winbase", "winnt"] }
 
 [dev-dependencies]
+futures = "0.3"
 serde_json = "1.0.39"
 tempfile = "3.2.0"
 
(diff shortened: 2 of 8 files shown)


0xpr03 (changes requested):
> but all the backends either guarantee only a single background thread, or they already use an Arc<Mutex<...>> to protect EventFn

That looks fine but can/do we somehow guarantee/note that we now expect only a single Fn caller at any point in time in the backends?

0xpr03:
Also thanks for all the changes. Based on your contributions, do you maybe want to be part of the team to maintain this crate, especially since you seem to have an apple device to test the fsevent part ?

erickt:
> That looks fine but can/do we somehow guarantee/note that we now expect only a single Fn caller at any point in time in the backends?

That should be implied by using `FnMut`, which will guarantee only one caller unless we make a mistake with `unsafe` and alias the function pointer. I looked through our use of `EventFn`, and it looks like all watchers but `INotifyWatcher` wrap their `event_fn` in `Arc<Mutex<...>>`. `INotifyWatcher` can get away with it only ever uses one background thread, so it can transfer ownership of the `EventFn` into it. We could change all the other watchers to use this pattern, which I think would allow us to get rid of the mutex.

> Also thanks for all the changes. Based on your contributions, do you maybe want to be part of the team to maintain this crate, especially since you seem to have an apple device to test the fsevent part ?

Thanks for the offer! I can try to help here and there, but I don't think I can commit to maintenance at this time. I think with these last couple changes addresses all my needs, and I can unblock my actual use of notify :)

0xpr03:
> We could change all the other watchers to use this pattern, which I think would allow us to get rid of the mutex.

I think we'll leave that for a later PR and merge this one in the meantime.

> I think with these last couple changes addresses all my needs, and I can unblock my actual use of notify 

No problem, let's see when your next issue comes up ;)

0xpr03 (approved):

PR #334, opened 2021-06-08
Tweak workspace structure

JohnTitor:
Followup of #332. With this, we can specify the example with the shorter name.

### DIFF
--- a/examples/hot_reload_tide/Cargo.toml
+++ b/examples/hot_reload_tide/Cargo.toml
@@ -11,4 +11,4 @@ tide = "0.16.0"
 async-std = { version = "1.6.0", features = ["attributes"] }
 serde_json = "1.0"
 serde = "1.0.115"
-notify = { path = "../..", features = ["serde"] }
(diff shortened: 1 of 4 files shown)


examples/hot_reload_tide/Cargo.toml
```diff
@@ -11,4 +11,4 @@ tide = "0.16.0"
 async-std = { version = "1.6.0", features = ["attributes"] }
 serde_json = "1.0"
 serde = "1.0.115"
-notify = { path = "../..", features = ["serde"] }
+notify = { version = "5.0.0-pre.10", features = ["serde"] }
```
0xpr03, line 14: Can you add this also to the release_checklist.md ? Because we've to change/check this one too.

0xpr03 (changes requested):

examples/hot_reload_tide/Cargo.toml
JohnTitor: Sure, added.

Issue #327, opened 2021-06-04
Proposal: kqueue support based on polling crate 


xanderio:
I'm currently looking into adding support for kqueue to support BSD platforms.

After a bit of digging i found the [polling](https://crates.io/crates/polling) crate from the smol-rs team.
The crate it self lists macOS, iOS, FreeBSD, NetBSD, OpenBSD, and DragonFly BSD as supported platforms for kqueue.  

Would PR adding polling based KqueueWatcher be accepted?

0xpr03:
Yep, see #273 (and also (here)[https://github.com/notify-rs/notify/issues?q=is%3Aissue+is%3Aopen+kqueue] for all the kqueue mentions about better support through that ;) )

0xpr03:
A bit of warning though: The current structure/implementation for the other systems is a little bit blurry as some stuff got changed/removed without a real new replacement. See for example [here](https://github.com/notify-rs/notify/pull/286#issuecomment-821239651). It's totally fine if you just copy the feature parity of the inotify implementation (for which I can speak) and not implement for example the precise config optionality. Still I'd welcome some hints in the code for `unsafe` for why it's safe and where you'd have to flip something to reach/implement unprecise* events.

\* With unprecise I mean the [PreciseEvents](https://docs.rs/notify/5.0.0-pre.9/notify/enum.Config.html#variant.PreciseEvents) option we want to reach.

Issue #136, opened 2017-12-28
Use kqueue native APIs directly to support i686 / macOS


passcod:
The kqueue binding currently doesn't build at all on i686 and variants, because it doesn't use the right types when doing the binding for those architectures. However, the kqueue crate seems abandoned. Given this is a pretty small API, it shouldn't be too complex to do it ourselves instead.

See failed Travis builds on i686-unknown-freebsd, for example this one: https://travis-ci.org/passcod/notify/jobs/322424840

You can also see the build failure locally by running `ci/cross-tests.sh build i686-unknown-freebsd`.

#### Want to help?

Here is the documentation for the wrapper we currently use: https://docs.worrbase.com/rust/kqueue/
That should also give you its source.

kqueue is documented in its manual page: https://www.freebsd.org/cgi/man.cgi?kqueue(2)

To test the backend, run `cargo test -p notify-backend-kqueue` in BSD.

If you don't have a BSD on hand, you can get one running by installing [Vagrant](https://www.vagrantup.com), checking out the `next` branch of this repo, and running `vagrant up`. That will bring up a FreeBSD11 virtual machine that you can then `vagrant ssh` into for developing. Use `vagrant rsync` or `vagrant rsync-auto` to sync up files to the virtual machine. However, note that this will bring up an `x86_64` machine. To test on i686 you'll have to set up another one.

You don't _have_ to reimplement the binding. If you find a kqueue binding that does what we want and is maintained, you can use that!

The goal here is to make Notify have a working backend on BSD on both x64 and x86. A bonus goal would be to support [kqueue on macOS](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html#//apple_ref/doc/uid/TP40005289-CH5-SW2), as well.

passcod:
Looks like this will be easier to achieve by using the mio kqueue implementation directly.

PR #335, opened 2021-06-09
Kqueue backend for use on BSD

xanderio:
Kqueue Backend for all the BSDs 🎉 

It's based on the kqueue crate and the in watcher implementation is based on the `INotifyWatcher`.

BSDs where this PR is check that it compiles: 

- [x] FreeBSD
- [ ] OpenBSD
- [ ] NetBSD
- [ ] DragonflyBSD

This relates to #327 and maybe #136

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,10 +22,10 @@ jobs:
           - stable
           - nightly
         os: [ubuntu-latest, macos-latest, windows-latest]
-        feature: []
+        features: ["serde"]
         include:
           - os: macos-latest
-            feature: "macos_kqueue"
+            features: "serde, macos_kqueue"
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -40,9 +40,8 @@ jobs:
 
       - name: check build
         env:
-          FEATURE: ${{ matrix.feature }}
         if: matrix.version != 'stable'
-        run: cargo check --features "serde $FEATURE" --examples
+        run: cargo check --features ${{ matrix.features }} --examples
 
       - name: check build hot reload tide
         if: matrix.version != 'stable'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,6 +22,10 @@ jobs:
           - stable
           - nightly
         os: [ubuntu-latest, macos-latest, windows-latest]
+        feature: []
+        include:
+          - os: macos-latest
+            feature: "macos_kqueue"
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -35,8 +39,10 @@ jobs:
           rustup override set ${{ matrix.version }}
 
       - name: check build
+        env:
+          FEATURE: ${{ matrix.feature }}
         if: matrix.version != 'stable'
-        run: cargo check --features serde --examples
+        run: cargo check --features "serde $FEATURE" --examples
 
       - name: test
         if: matrix.version == 'stable'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,10 +22,6 @@ jobs:
           - stable
           - nightly
         os: [ubuntu-latest, macos-latest, windows-latest]
-        features: ["serde"]
-        include:
-          - os: macos-latest
-            features: "serde, macos_kqueue"
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -39,8 +35,12 @@ jobs:
           rustup override set ${{ matrix.version }}
 
       - name: check build
-        if: matrix.version != 'stable'
-        run: cargo check --features ${{ matrix.features }} --examples
+        if: matrix.version != 'stable' && matrix.os == 'macos-latest'
+        run: cargo check --features=serde,macos_kqueue --examples
+
+      - name: check build
+        if: matrix.version != 'stable' && matrix.os != 'macos-latest'
+        run: cargo check --features=serde --examples
 
       - name: check build hot reload tide
         if: matrix.version != 'stable'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,6 @@ jobs:
           rustup override set ${{ matrix.version }}
 
       - name: check build
-        env:
         if: matrix.version != 'stable'
         run: cargo check --features ${{ matrix.features }} --examples
 
--- a/src/kqueue.rs
+++ b/src/kqueue.rs
@@ -5,7 +5,7 @@
 //! pieces of kernel code termed filters.
 
 use super::event::*;
-use super::{Config, Error, EventFn, RecursiveMode, Result, Watcher};
+use super::{Error, EventFn, RecursiveMode, Result, Watcher};
 use crossbeam_channel::{unbounded, Sender};
 use kqueue::{EventData, EventFilter, FilterFlag, Ident};
 use std::collections::HashMap;
--- a/src/kqueue.rs
+++ b/src/kqueue.rs
@@ -163,7 +163,7 @@ impl EventLoop {
                                     Event::new(EventKind::Remove(RemoveKind::Any))
                                 }
 
-                                //data was write to this file
+                                //data was written to this file
                                 kqueue::Vnode::Write => Event::new(EventKind::Access(
                                     AccessKind::Close(AccessMode::Write),
                                 )),
@@ -190,7 +190,7 @@ impl EventLoop {
 
                                 /*
                                 The link count on a file changed => subdirectory created or
-                                delete. Currently now idea how to track this.
+                                delete. Currently no idea how to track this.
                                 */
                                 kqueue::Vnode::Link => {
                                     // readd this folder, this is currently the easiest way to
--- a/src/kqueue.rs
+++ b/src/kqueue.rs
@@ -1,0 +1,287 @@
+//! Watcher implementation for the inotify Linux API
+//!
+//! The inotify API provides a mechanism for monitoring filesystem events.  Inotify can be used to
+//! monitor individual files, or to monitor directories.  When a directory is monitored, inotify
+//! will return events for the directory itself, and for files inside the directory.
+
+use super::event::*;
+use super::{Config, Error, EventFn, RecursiveMode, Result, Watcher};
+use crossbeam_channel::{unbounded, Sender};
+use kqueue::{EventFilter, FilterFlag};
+use std::collections::HashMap;
+use std::env;
+use std::fs::metadata;
+use std::os::unix::io::AsRawFd;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use std::thread;
+use walkdir::WalkDir;
+
+const KQUEUE: mio::Token = mio::Token(0);
+const MESSAGE: mio::Token = mio::Token(1);
+
+// The EventLoop will set up a mio::Poll and use it to wait for the following:
+//
+// -  messages telling it what to do
+//
+// -  events telling it that something has happened on one of the watched files.
+struct EventLoop {
+    running: bool,
+    poll: mio::Poll,
+    event_loop_waker: Arc<mio::Waker>,
+    event_loop_tx: crossbeam_channel::Sender<EventLoopMsg>,
+    event_loop_rx: crossbeam_channel::Receiver<EventLoopMsg>,
+    kqueue: kqueue::Watcher,
+    event_fn: Box<dyn EventFn>,
+    watches: HashMap<PathBuf, bool>,
+}
+
+/// Watcher implementation based on inotify
+pub struct KqueueWatcher {
+    channel: crossbeam_channel::Sender<EventLoopMsg>,
+    waker: Arc<mio::Waker>,
+}
+
+enum EventLoopMsg {
+    AddWatch(PathBuf, RecursiveMode, Sender<Result<()>>),
+    RemoveWatch(PathBuf, Sender<Result<()>>),
+}
+
+impl EventLoop {
+    pub fn new(kqueue: kqueue::Watcher, event_fn: Box<dyn EventFn>) -> Result<Self> {
+        let (event_loop_tx, event_loop_rx) = crossbeam_channel::unbounded::<EventLoopMsg>();
+        let poll = mio::Poll::new()?;
+
+        let event_loop_waker = Arc::new(mio::Waker::new(poll.registry(), MESSAGE)?);
+
+        let kqueue_fd = kqueue.as_raw_fd();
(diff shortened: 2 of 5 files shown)


0xpr03:
I enabled the CI for this PR so if possible you can setup & test github CI for freebsd

xanderio:
Thanks, but Github Actions don't have support for FreeBSD.

I found this action[1], but i'm not sure if this is a greate solution for that problem.

[1]: https://github.com/vmactions/freebsd-vm

xanderio:
We could setup a ci pipeline under MacOS in combination with a feature to select between FSEvent and kqueue.

JohnTitor:
FWIW, the libc crate and Tokio use Cirrus CI (e.g. https://github.com/tokio-rs/tokio/blob/master/.cirrus.yml) for FreeBSD testing though it has some limitations for free users.

xanderio:
Which solution should i implement? 

1. Cargo feature to toggle the mac backend to kqueue
2. Cirrus CI on FreeBSD

xanderio:
The only thing the I'm currently seeing that is still missing is some kind of CI solution.

JohnTitor:
I don't have a strong opinion here but is the first option beneficial for mac users as well? If so, it's better than the second as we don't need additional CI.

0xpr03:
I think we want some freebsd CI if possible. I'm fine with testing this only on mac if nothing else, but to have official BSD support it'd be very nice to have a CI run. Otherwise you have a very high chance it'll break and nobody will notice. @JohnTitor can you help setup cirrus CI for that if possible ?

JohnTitor:
> Otherwise you have a very high chance it'll break and nobody will notice.

AFAIK macOS is one of BSD targets and it should have kqueue API. Does this have FreeBSD-specific operations?

I'm happy to help setup Cirrus CI though :)

worr:
Hey, original author of the `kqueue` crate here.

This doesn't have anything FreeBSD-specific, since as of right now, the high-level parts of the kqueue are all cross-platform. I've looked at the NOTEs and stuff they're using and ensured that they're also cross-platform. In other words, LGTM.

While macOS does support `kqueue(2)` and theoretically this could be used there, not sure what cases someone would want to use `kqueue(2)` rather than `FSEvents` which is more native, and afaict already supported by this crate. A feature like @xanderio suggested might be a good way to gate it, for either easier CI or in case someone v specifically wants to use `kqueue(2)`. That's just my 2¢ tho, feel free to disregard 🙂

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+                }
+            }
+        }
+    }
+
+    fn handle_kqueue(&mut self) {
+        let mut add_watches = Vec::new();
+        let mut remove_watches = Vec::new();
+
+        println!("hello");
```
0xpr03, line 142: I think this should be removed before merging

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+/// return `DirEntry` when it is a directory
+fn filter_dir(e: walkdir::Result<walkdir::DirEntry>) -> Option<walkdir::DirEntry> {
+    if let Ok(e) = e {
+        if let Ok(metadata) = e.metadata() {
```
0xpr03, line 308: The caller functions returns an error, so I'd like to report permission/access errors here too and not silently drop them

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+        let pb = if path.is_absolute() {
+            path.to_owned()
+        } else {
+            let p = env::current_dir().map_err(Error::io)?;
+            p.join(path)
+        };
+        let (tx, rx) = unbounded();
+        let msg = EventLoopMsg::AddWatch(pb, recursive_mode, tx);
+
+        // we expect the event loop to live and reply => unwraps must not panic
```
0xpr03, line 337: I think we could also just report this via the Result and not panic randomly inside application code. But that's not a hard opinion, just an idea.

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+        let pb = if path.is_absolute() {
+            path.to_owned()
+        } else {
+            let p = env::current_dir().map_err(Error::io)?;
+            p.join(path)
+        };
+        let (tx, rx) = unbounded();
+        let msg = EventLoopMsg::RemoveWatch(pb, tx);
+
+        // we expect the event loop to live and reply => unwraps must not panic
```
0xpr03, line 353: same goes here if applicable

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+                                could only emulate this inotify behavior if we keep track of
+                                all files and directories internally and then perform a
+                                lookup.
+                                */
+                                kqueue::Vnode::Delete => {
+                                    remove_watches.push(path.clone());
+                                    Event::new(EventKind::Remove(RemoveKind::Any))
+                                }
+
+                                //data was write to this file
```
0xpr03, line 166: *written

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+                                kqueue::Vnode::Write => Event::new(EventKind::Access(
+                                    AccessKind::Close(AccessMode::Write),
+                                )),
+
+                                /*
+                                Extend and Truncate are just different names for the same
+                                operation, extend is only used on FreeBSD, truncate everwhere
+                                else
+                                */
+                                kqueue::Vnode::Extend => Event::new(EventKind::Modify(
```
0xpr03, line 176: You can change this to `kqueue::Vnode::Extend | kqueue::Vnode::Truncate =>` ..

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+                                able of delete. In this case it would somewhat expensive to
+                                keep track and compare ever peace of metadata for every file
+                                */
+                                kqueue::Vnode::Attrib => Event::new(EventKind::Modify(
+                                    ModifyKind::Metadata(MetadataKind::Any),
+                                )),
+
+                                /*
+                                The link count on a file changed => subdirectory created or
+                                delete. Currently now idea how to track this.
```
0xpr03, line 196: *no idea

src/kqueue.rs
```diff
@@ -0,0 +1,380 @@
...
+                                */
+                                kqueue::Vnode::Link => {
+                                    // readd this folder, this is currently the easiest way to
+                                    // track new and removed subfolders
+
+                                    // TODO: This is really expensive, as we recursively walk through
+                                    // all subdirectories.
+                                    remove_watches.push(path.clone());
+                                    add_watches.push(path.clone());
```
0xpr03, line 205: @worr do you know of a better way than to re-index the whole tree ?

0xpr03 (changes requested):
Thanks for the PR. I've added some minor questions or notes what could be improved. If possible I'd like to resolve the todos. I think we can also just track some stuff if we're already re-indexing the whole path upon link change.

src/kqueue.rs
xanderio: This code is copied directly from the INotifyWatcher implementation. So we should resolve this also in there

src/kqueue.rs
xanderio: This is also copied from the INotifyWatcher

src/kqueue.rs
xanderio: This is also copied the INotifyWatcher

src/kqueue.rs
xanderio: We could store a HashMap of all know paths and then compare with paths are missing or where added to the directory.
I'll try to implement this.

xanderio: After thinking about this a bit, i cam to the conclusion that a simple `HashSet<PathBuf>` won't cut it. As i don't think there is any way to get from the directory we get from `NOTE_LINK` to the files we expect to be in that directory.

I think we need a `HashMap<PathBuf, Vec<PathBuf>>` where the key is the directory and the value is a list of all files in the directory at time of adding. This feels like a race condition to me, as files could be created of deleted will we are setting up this directory watch. Not quite sure if this is important in this case.

src/kqueue.rs
xanderio: After looking over this peace of code, i agree with you. To which `ErrorKind` should i map this, or should this maybe be resolved in an another  PR where this behavior is also removed from `INotifyWatcher`.

src/kqueue.rs
xanderio: After looking over this peace of code, i agree with you. To which `ErrorKind` should i map this, or should this maybe be resolved in an another PR where this behavior is also removed from `INotifyWatcher`.

src/kqueue.rs
worr: @0xpr03 We've been chatting about this out-of-band a bit. `kqueue(2)` doesn't expose this in a meaningful way, and this is especially expensive since we need to open file handles to each file on index (this is a requirement of `kqueue(2)`). The method @xanderio brought up is one of the ones we've been talking about.

0xpr03: Hm, thanks for investigating. I'm fine with the current solution, it's not the first time where we have to re-index due to race conditions. Could you change the todo so it documents why we have to use this approach ?

src/kqueue.rs
0xpr03: error::Generic as also requested [here](https://github.com/notify-rs/notify/pull/337#discussion_r652165310) for anther in-flight PR

src/kqueue.rs
0xpr03: error::Generic is fine

0xpr03:
Ping me again here if you're not hearing anything from me, I just discovered your comments on the review by accident and didn't get any notification. (Also the github CI run apparently has to be re-approved every time again, as long as you haven't made at least one commit into the repo.)

src/kqueue.rs
xanderio: I'll can do that, all though I personally would prefer if  we left the TODO in places where there are thinks that we see that can possibly be improve in the future, but we just don't know the solution to at the moment. But that's just my 2ct

xanderio:
Quick side note: After this is merged could you release a new preview build?

0xpr03:
Yeah, I'll probably merge #337 first and then release them together when I cleaned up the changelog, some stuff got left out.

xanderio:
After we fix the could of code review issues with this pr. The only thing that is missing is some kind of CI setup. 
I would start working on the macos based solution previously discussed, if i don't here anything else from you.

0xpr03:
I'm sorry but I'm not exactly sure what you mean with that. If you want you can go ahead and make a macOS PR.

src/kqueue.rs
0xpr03: on a second thought yeah, you can leave it like that

xanderio:
I mean adding a cargo `macos_kqueue` feature which enables the `kqueue` backend for macos. With that we could add CI to this PR without having to figure out how to setup cirrus ci for an opensource project.

0xpr03:
Yeah that sounds like a good idea. @JohnTitor do you prefer doing this via cirrus or just a cargo feature ?

xanderio:
I would like to leave the  `TODO` on the `NOTE_DELETE` in the source code, because this is a valid improvement that can be made in the future.

src/kqueue.rs
```diff
@@ -0,0 +1,389 @@
...
+//!
+//! The kqueue() system call provides a generic method of notifying the user
+//! when an event happens or a condition holds, based on the results of small
+//! pieces of kernel code termed filters.
+
+use super::event::*;
+use super::{Error, EventFn, RecursiveMode, Result, Watcher};
+use crossbeam_channel::{unbounded, Sender};
+use kqueue::{EventData, EventFilter, FilterFlag, Ident};
+use std::collections::{HashMap, HashSet};
```
0xpr03, line 11: HashSet should be irrelevant then

0xpr03 (changes requested):

0xpr03:
> 
> 
> I would like to leave the `TODO` on the `NOTE_DELETE` in the source code, because this is a valid improvement that can be made in the future.

ok

xanderio:
The latest commit include a CI pipeline, as this is the first time i ever did something with github actions, I'm not sure if this is correct. 

Other then the unwrap issue on line 343 and 359. This PR is in my eyes almost complete. 🎉

.github/workflows/main.yml
```diff
@@ -35,8 +39,10 @@ jobs:
           rustup override set ${{ matrix.version }}
 
       - name: check build
+        env:
+          FEATURE: ${{ matrix.feature }}
         if: matrix.version != 'stable'
-        run: cargo check --features serde --examples
+        run: cargo check --features "serde $FEATURE" --examples
```
JohnTitor, line 45: IIRC we can use `${{ matrix.feature }}` directly.

JohnTitor:
> Yeah that sounds like a good idea. @JohnTitor do you prefer doing this via cirrus or just a cargo feature ?

I don't have a strong opinion but it's worth trying out the latter if we've implemented it.

.github/workflows/main.yml
```diff
@@ -22,6 +22,10 @@ jobs:
           - stable
           - nightly
         os: [ubuntu-latest, macos-latest, windows-latest]
+        feature: []
```
JohnTitor, line 25: > The workflow is not valid. .github/workflows/main.yml (Line: 25, Col: 18): Matrix vector 'feature' does not contain any values

The default could be "serde" so that we can simplify the below command on `check build`.

JohnTitor: nit: And it should be `features`.

xanderio:
Just pushed to last commit, which should ™️  fix all code review issus

.github/workflows/main.yml
```diff
@@ -35,8 +39,9 @@ jobs:
           rustup override set ${{ matrix.version }}
 
       - name: check build
+        env:
```
JohnTitor, line 42:
```suggestion
```

xanderio:
Thanks for fixing the CI @JohnTitor 

@0xpr03 According to github you're still requesting changes, but i can't figure out what changes those are. :/

0xpr03 (approved):

0xpr03:
I think that was simply because I had a change-request review with a [general comment](https://github.com/notify-rs/notify/pull/335#pullrequestreview-683781485) which probably isn't resolved by the individual code comments.

xanderio:
I'm wondering if there is something that this PR is waiting on or if you just haven't come around to merge it?

0xpr03:
Thanks, was busy last week and githubs UI made it looks like merging this isn't possible without manual merge conflict intervention.

PR #336, opened 2021-06-10
Make `Watcher` object safe

erickt:
This changes the `Watcher` trait to be object safe, which allows users to
dynamically select which backend to use. This can be helpful on systems that
support multiple file watching systems that have different tradeoffs. For
example, Chromium's file watcher on OS X will use an fsevent backend
when watching recursive directories, and a kqueues backend when not.

In order to implement this, this makes a few changes to Watcher:

* removes `new_immediate` constructor.
* removes the `<P: AsRef<Path>>` from the `watch` and `unwatch`.

This then replaces (and renames) calls with `notify::recommended_watcher()`
to get similar behavior as the old constructor.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -499,12 +499,12 @@ impl ReadDirectoryChangesWatcher {
 }
 
 impl Watcher for ReadDirectoryChangesWatcher {
-    fn watch<P: AsRef<Path>>(&mut self, path: P, recursive_mode: RecursiveMode) -> Result<()> {
-        self.watch_inner(path.as_ref(), recursive_mode)
+    fn watch(&mut self, path: &Path, recursive_mode: RecursiveMode) -> Result<()> {
+        self.watch_inner(path, recursive_mode)
     }
 
-    fn unwatch<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
-        self.unwatch_inner(path.as_ref())
+    fn unwatch(&mut self, path: &Path) -> Result<()> {
+        self.unwatch_inner(path)
     }
 
     fn configure(&mut self, config: Config) -> Result<bool> {
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -403,6 +403,14 @@ pub struct ReadDirectoryChangesWatcher {
 }
 
 impl ReadDirectoryChangesWatcher {
+    pub fn new<F: EventFn>(event_fn: F) -> Result<Self> {
+        // create dummy channel for meta event
+        // TODO: determine the original purpose of this - can we remove it?
+        let (meta_tx, _) = unbounded();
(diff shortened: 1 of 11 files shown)


0xpr03:
I think`<P: AsRef<Path>>` was nice to leave freedom in how you want to specify your path. Any reason to remove that ?

src/windows.rs
```diff
@@ -403,6 +403,14 @@ pub struct ReadDirectoryChangesWatcher {
 }
 
 impl ReadDirectoryChangesWatcher {
+    pub fn new<F: EventFn>(event_fn: F) -> Result<Self> {
+        // create dummy channel for meta event
+        // TODO: determine the original purpose of this - can we remove it?
```
0xpr03, line 408: I think this is from the original implementation for debouncing where each Watcher also implemented something like "new debounced" (though it was swapped, new was debounced) and setup the channel system internally.

Problem is that we have to pass-through events from the watcher to the debouncer to the user but the user received a debouncer object that has to pass configuration changes back to the watcher. (Which is why I've currently halted adding a new debouncer until I've got a better API idea.)

0xpr03: And the same happens but with a raw watcher object without any debouncer, which makes this a really bad data chain.

erickt: note I just copied this comment over from the old `ReadDirectoryChangesWatcher::new_immediate`, so I don't have context why this comment was added :)

erickt:
> I think<P: AsRef<Path>> was nice to leave freedom in how you want to specify your path. Any reason to remove that ? I guess we have to remove that to allow dynamic switching via non-generic interfaces.

Correct, object safe traits can't have generic parameters. It's unfortunate to have to manually pass in `Path::new()`, but I'm guessing most callsites only specify one or two paths by hand, so it might not be that painful in practice.

0xpr03:
Note that you currently seem to have syntax errors according to CI.

0xpr03 (changes requested):
Note that you currently seem to have syntax errors according to CI.

erickt:
> Note that you currently seem to have syntax errors according to CI.

Should be fixed now!

0xpr03 (approved):

0xpr03:
Thanks, this will be a good step towards a possible dynamic runtime change, which was in the never released -next branch

0xpr03:
Note: we may want to document and test for this, so it isn't reverted later on with a trait under the impression of usability improvements

erickt:
Thanks!

> Note: we may want to document and test for this, so it isn't reverted later on with a trait under the impression of usability improvements

I did add a simple test for this in src/lib.rs, where I assert that the `Watcher` trait is object safe. So we can't accidentally remove this feature.

PR #337, opened 2021-06-10
fsevent should join on thread shutdown

erickt:
We had some test failures because crossbeam-channel may panic when trying to call recv() during thread shutdown in a drop(). This seems to be similar to this upstream bug: https://github.com/crossbeam-rs/crossbeam/issues/321. This appears to be happening because recv() interacts with thread local storage, but unfortunately it seems that some operating systems may tear down thread-local storage early, https://github.com/rust-lang/rust/issues/28129, which can trigger panics if trying to interact with TLS during a drop.

To avoid this issue, this switches from using a channel to signal the thread shutdown to just using the join handle (which we should have been doing anyway).

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -16,7 +16,7 @@
 
 use crate::event::*;
 use crate::{Config, Error, EventFn, RecursiveMode, Result, Watcher};
-use crossbeam_channel::{unbounded, Receiver, Sender};
+use crossbeam_channel::{unbounded, Sender};
 use fsevent_sys as fs;
 use fsevent_sys::core_foundation as cf;
 use std::collections::HashMap;
@@ -65,7 +65,7 @@ pub struct FsEventWatcher {
     latency: cf::CFTimeInterval,
     flags: fs::FSEventStreamCreateFlags,
     event_fn: Arc<Mutex<dyn EventFn>>,
-    runloop: Option<(cf::CFRunLoopRef, Receiver<()>)>,
+    runloop: Option<(cf::CFRunLoopRef, thread::JoinHandle<()>)>,
     recursive_info: HashMap<PathBuf, bool>,
 }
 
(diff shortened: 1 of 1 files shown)


erickt:
Also, I haven't yet figured out a workaround for this issue on Macs, so would it be possible to cut a release once this lands?

src/fsevent.rs
```diff
@@ -450,13 +444,9 @@ impl FsEventWatcher {
...
                 fs::FSEventStreamRelease(stream);
             }
-
-            done_tx
-                .send(())
-                .expect("error while signal run loop is done");
         });
         // block until runloop has been sent
-        self.runloop = Some((rl_rx.recv().unwrap().0, done_rx));
+        self.runloop = Some((rl_rx.recv().unwrap().0, thread_handle));
```
0xpr03, line 449: Is there a chance this might fail / get poisoned ? Because then I'd like to not just panic and use the Error we can return.

0xpr03 (approved):

src/fsevent.rs
erickt: I think the only place where the thread will panic would be when we're trying to send the runloop back to the main thread. I don't believe that can happen here, as best as I can tell. Furthermore, the old behavior would have also panicked in `stop()` if we didn't receive confirmation we shut down the thread.

We could update this to propagate the PoisonError though. Would you want this converted into error::Error::Generic, or to add a new error variant?

0xpr03: I think error::Error::Generic is totally fine. I'd just try to have less code/crates that can panic your application instead of catching such errors at points the developer can decide.

0xpr03: ping: do you want to resolve this or should I go on, merge it and issue a release so it can be fixed later ?

erickt: Hello! Sorry, got distracted. Anyway, looking back through the code there's no place where the code could panic in our code. So the only place this that could happen would be if the fsevents crate (which is unlikely since we're just making direct calls into the C library) or std::thread library panicked. Since we can't be assured what internal invariants might have been violated if that happens, I think panicking is the appropriate course of action.

PR #345, opened 2021-08-05
Move creation of watcher into trait

0xpr03:
I don't know why we haven't done this earlier, but I think unifying the creation of the watcher by moving it into the trait shouldn't be a problem, if I'm not missing some lifetime problem.


JohnTitor (approved):
Seems `KqueueWatcher` has been slipped but otherwise LGTM 👍

PR #346, opened 2021-08-06
Add `EventHandler` trait to replace `EventFn`

a1phyr:
Replace `EventFn`trait with `EventHandler`.

The new trait has new capabilities but can be implemented for more types, such as channel receivers which are very commonly used.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,6 +139,23 @@ mod config;
 mod error;
 
 /// The set of requirements for watcher event handling functions.
+///
+/// # Example implementation
+///
+/// ```no_run
+/// use notify::{Event, Result, EventHandler};
+///
+/// /// Prints received events
+/// struct EventPrinter;
+///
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -139,9 +139,31 @@ mod config;
 mod error;
 
 /// The set of requirements for watcher event handling functions.
```
0xpr03, line 141: Can you add an example in the documentation for using the new impls ?

0xpr03 (changes requested):
Hi, can you rebase this onto the current master ? Other than that it seems fine, except for more documentation.

a1phyr:
Done

0xpr03:
Thanks! I'll add an example to the lib file. Maybe it'll be worth wile to add a tokio feature where we also implement this for async tokio channels. Problem is though, that we'll have to select between tokio::sync::broadcast and tokio::sync::watch or tokio::sync::mpsc.

PR #350, opened 2021-08-13
[NO MERGE] test build failure

0xpr03:


0xpr03:
Just for confirmation @lnicola it's apparently working with freebsd but fails on netbsd, so that means the 1.0.4 should actually fix it.

PR #352, opened 2021-08-21
Use a helper to check a modified event or not

y-yagi:
I tried this example but the config didn't reload when that was changed.
I think the original check didn't work as expected, so I fixed to use a helper function.

### DIFF
--- a/examples/hot_reload_tide/src/main.rs
+++ b/examples/hot_reload_tide/src/main.rs
@@ -3,9 +3,7 @@
 // you can edit the configuration and the app will pick up changes without the need to restart it.
 // This concept is known as hot-reloading.
 use hot_reload_tide::messages::{load_config, Config};
-use notify::{
-    event::ModifyKind, Error, Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher,
-};
+use notify::{Error, Event, RecommendedWatcher, RecursiveMode, Watcher};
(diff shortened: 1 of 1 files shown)


0xpr03:
Yeah good catch. It would have to match EventKind::Modify(_), which the function call does. Otherwise you'll not match precise events.

examples/hot_reload_tide/src/main.rs
```diff
@@ -3,9 +3,7 @@
 // you can edit the configuration and the app will pick up changes without the need to restart it.
 // This concept is known as hot-reloading.
 use hot_reload_tide::messages::{load_config, Config};
-use notify::{
-    event::ModifyKind, Error, Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher,
-};
+use notify::{Error, Event, RecommendedWatcher, RecursiveMode, Watcher};
```
0xpr03, line 6: Is that done manually ? Because I think this was from rust fmt.

0xpr03 (approved):

examples/hot_reload_tide/src/main.rs
y-yagi: No, I did this by `cargo fmt`. 
I think this is because the length of this line was changed less than the value of `max_width`(default value is 100).

PR #360, opened 2021-09-18
PollWatcher: unify signature of `new` and `with_delay`

jhscheer:
* Move `event_handler` logic from `new` to `with_delay` to be able to call:
```rust
let (tx, rx) = std::sync::mpsc::channel();
let mut watcher = PollWatcher::with_delay(tx, Duration::from_secs(1))?;
```
instead of:
```rust
let (tx, rx) = std::sync::mpsc::channel();
let mut watcher = PollWatcher::with_delay(Arc::new(Mutex::new(tx)), Duration::from_secs(1))?;
```

* Add documentation

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -43,13 +43,10 @@ fn emit_event(event_handler: &Mutex<dyn EventHandler>, res: Result<Event>) {
 }
 
 impl PollWatcher {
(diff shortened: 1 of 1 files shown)


0xpr03 (approved):
Thanks, I thought a bit about whether we may want to allow passing an Arc inside, but ultimately we already do the same thing for other backends..

PR #362, opened 2021-09-29
Fix selection of `RecommendedWatcher` for `macos_kqueue` feature.

jhscheer:
This fixes an issue on macOS where `RecommendedWatcher` is `FsEventWatcher`
although `features="macos_kqueue"` is selected at compile  time.


0xpr03 (approved):
Thanks

PR #364, opened 2021-09-29
Add Watcher::kind()

### LINKED ISSUE #361: Which Watcher do I get?
I would like to know which Watcher `RecommendedWatcher::new()` returns.

For example if `let mut watcher = RecommendedWatcher::new(tx)?;` returns a `PollWatcher` I would like to set the delay manually (the default is 30 seconds and cannot be adjusted later).

Maybe something like,

1. either add an `Option<Duration>` parameter to `Watcher::new()` which will be relevant only if `type RecommendedWatcher = PollWatcher;` Then I could call `let mut watcher = RecommendedWatcher::new(tx, Some(Duration::from_secs(1)))?;`
2. or add a `whoami()` method to the Watcher trait and implement setting the delay with `configure()`. Then I could call, e.g. 
```rust
let mut watcher = RecommendedWatcher::new(tx)?;
if watcher.whoami == WatcherType::PollWatcher {
    watcher.configure(Config::SetDelay(Duration::from_secs(1)))?;
}

### PR DESCRIPTION
**Pull request by 0xpr03**
Closes #361

### DIFF
--- a/examples/watcher_kind/src/main.rs
+++ b/examples/watcher_kind/src/main.rs
@@ -1,0 +1,11 @@
+use std::time::Duration;
+
+use notify::*;
+fn main() {
+    let (tx, rx) = std::sync::mpsc::channel();
+    let watcher: Box<dyn Watcher> = if RecommendedWatcher::kind() == WatcherKind::PollWatcher {
+        Box::new(PollWatcher::with_delay(tx,Duration::from_secs(1)).unwrap())
+    } else {
+        Box::new(RecommendedWatcher::new(tx).unwrap())
+    };
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -185,6 +185,24 @@ impl EventHandler for std::sync::mpsc::Sender<Result<Event>> {
     }
 }
 
+/// Watcher kind enumeration
+#[derive(Debug,PartialEq,Eq)]
+#[non_exhaustive]
+pub enum WatcherKind {
+    /// inotify backend (linux)
+    Inotify,
+    /// FS-Event backend (mac)
+    Fsevent,
+    /// KQueue backend (bsd,mac)
+    Kqueue,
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by 0xpr03** [comment]
cc @jhscheer would that be an appropriate fix for you ?

**Comment by jhscheer** [comment]
Yes, this looks very good and solves my issue. Thanks!

**Comment by 0xpr03** [comment]
Note that non-boxing would require an enum that matches the CFG chaos of RecommendedWatcher, adding an enum for every possible configurable watcher combination. And otherwise we'll have to add the specific config calls to the trait.

File: examples/watcher_kind/src/main.rs [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+    
+    let (tx, rx) = std::sync::mpsc::channel();
+    let watcher: Box<dyn Watcher> = if RecommendedWatcher::kind() == WatcherKind::PollWatcher {
+        Box::new(PollWatcher::with_delay(tx,Duration::from_secs(1)).unwrap())
+    } else {
+        Box::new(RecommendedWatcher::new(tx).unwrap())
+    };
+
+
+}
```
JohnTitor, line 14: It'd be great if we could remove blank lines (i.e. follow rustfmt).

src/lib.rs
```diff
@@ -185,6 +185,23 @@ impl EventHandler for std::sync::mpsc::Sender<Result<Event>> {
     }
 }
 
+/// Watcher kind enumeration
+#[derive(Debug,PartialEq,Eq)]
+pub enum WatcherKind {
```
JohnTitor, line 190: I guess we'll add/remove watchers in the future (actually, kqueue has been added recently) so it should make sense to add the `#[non_exhaustive]` attribute.

JohnTitor (approved):
Overall LGTM with a few nits 👍

0xpr03:
Should both be addressed

PR #369, opened 2021-11-20
fsevent: return error on canonicalize failure

### LINKED ISSUE #368: Panic in FSEvents when canonicalizing missing path
#### System details

- OS/Platform name and version: OSX / FSEvents
- Rust version (if building from source): `rustc --version`: `rustc 1.54.0 (a178d0322 2021-07-26)`
- Notify version (or commit hash if building from git): `v5.0.0-pre.13`

I have a unit test which tests the ability for some code to clear out file cache when done with its work. Notify is used to assert that the file(s) are removed:

```rust
pub async fn wait_for_path_removal(p: impl AsRef<Path>) {
    use notify::{EventKind, RecursiveMode, Watcher};
    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
    let _tx_guard = tx.clone();
    let mut w = notify::recommended_watcher(
        move |ev: std::result::Result<notify::Event, notify::Error>| {
            let ev = match ev {
                Ok(ev) => ev,
                Err(e) => {
                    log::error!("error in notify: {}", e);
                    return;
                }
            };
            if let EventKind::Remove(_rk) = ev.kind {
                tx.send(()).expect("Receiver was dropped!");
            }
        },
    )
    .unwrap();
    if let Err(e) = w.watch(p.as_ref(), RecursiveMode::NonRecursive) {
        match &e.kind {
            notify::ErrorKind::Io(io_err) if std::io::ErrorKind::NotFound == io_err.kind() => return,
            notify::ErrorKind::PathNotFound => return,
            _ => {}
        }
        panic!("Didn't match expected error: {}", e);
    };
    rx.recv().await.unwrap();
}
```

#### What you expected

The call to `Watcher::watch` should succeed if the file exists, or return an error so that the function may return.

#### What happened

A panic was hit when unwrapping the path `canonicalize` attempt:

```
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
thread 'test::add_asset_update' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /Users/jonmagnuson/.cargo/registry/src/github.com-1ecc6299db9ec823/notify-5.0.0-pre.13/src/fsevent.rs:364:47
```

https://github.com/notify-rs/notify/blob/b3f2d77e5c66e896b7a133fabd649b143bfe8b9d/src/fsevent.rs#L364

jmagnuson:
Removes a panic that can occur in `append_path` when a file is removed
after passing the `exists` check, but before canonicalization. This
change moves `canonicalize` before appending the path, so that it can
safely bail on error without leaking state.

Fixes #368.

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -347,6 +347,7 @@ impl FsEventWatcher {
         if !path.exists() {
             return Err(Error::path_not_found().add_path(path.into()));
         }
+        let canonical_path = path.to_path_buf().canonicalize()?;
         let str_path = path.to_str().unwrap();
         unsafe {
             let mut err: cf::CFErrorRef = ptr::null_mut();
@@ -361,7 +362,7 @@ impl FsEventWatcher {
             cf::CFRelease(cf_path);
         }
         self.recursive_info.insert(
-            path.to_path_buf().canonicalize().unwrap(),
+            canonical_path,
             recursive_mode.is_recursive(),
         );
         Ok(())


JohnTitor (approved):

src/fsevent.rs
```diff
@@ -347,6 +347,7 @@ impl FsEventWatcher {
         if !path.exists() {
             return Err(Error::path_not_found().add_path(path.into()));
         }
+        let canonical_path = path.to_path_buf().canonicalize()?;
```
0xpr03, line 350: Can we catch this here when it's `value: Os { code: 2, kind: NotFound, message: "No such file or directory" }`, so we don't emit anything in that case ? Would be cleaner I think.

0xpr03 (changes requested):
Thanks for the issue + PR, I think me want to catch this one case, otherwise it looks good.

src/fsevent.rs
jmagnuson: Are you proposing to panic on other `io::Error` variants? Unfortunately the [docs](https://doc.rust-lang.org/std/fs/fn.canonicalize.html#errors) are a bit vague about what other variants are possible, but there's also `NotADirectory` which could happen if the caller accidentally supplied a malformed path ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=036fd36ddcdcf3ce69475a893aed366c)). Granted, that example should likely result in a `path.exists()` bail, but there may be a pathological case I'm not thinking of which passes that but not `canonicalize`. 🤷 

The other route I had considered when implementing this was to mirror the behavior of `FsEventWatcher::remove_path`:

https://github.com/notify-rs/notify/blob/b3f2d77e5c66e896b7a133fabd649b143bfe8b9d/src/fsevent.rs#L334-L338

whereby the original, un-canonicalized path is used as a fallback if there's an error in `canonicalize`.

0xpr03: No, I was thinking about not emitting anything in that case and just aborting this event. But the more I think about it, the more it's probably the best if we let the user know about it. Even though it is a little bit confusing and most people probably aren't interested in knowing that something happened too fast to even record the exact event.

PR #370, opened 2021-11-23
kqueue: emit EventKind::Modify on kqueue write event

xanderio:
This should make the event emitted by the kqueue backend to be more in
line with the inotify backend and the documentation on `EventKind`.


0xpr03 (approved):

0xpr03:
thanks

PR #371, opened 2021-12-02
fsevent: use `RenameMode::Any` for renaming events

andelf:
> FSEvents provides no mechanism to associate the old and new sides of a rename event. It only produces an event at the old and new paths with `kFSEventStreamEventFlagItemRenamed` bit set.

Ref: https://github.com/atom/watcher/blob/master/docs/macos.md

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -172,9 +172,11 @@ fn translate_flags(flags: StreamFlags, precise: bool) -> Vec<Event> {
         });
     }
 
+    // FSEvents provides no mechanism to associate the old and new sides of a
+    // rename event.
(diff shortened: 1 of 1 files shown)


0xpr03:
Thanks for the PR. Can we somehow test this ? I don't use macos/bsd at all, so having some kind of sanity test for this would be very appreciated.

dhiemaz:
I use mac here.. @0xpr03

0xpr03:
@dhiemaz did you verify this ?

dhiemaz:
Let me check it first @0xpr03

dhiemaz:
Okay from me..

0xpr03 (approved):
Thanks!

PR #372, opened 2021-12-07
Kqueue emit Create Events and watch all files in a directory

### LINKED ISSUE #365: Create Event for kqueue missing
I'm using kqueue backend on macOS because FSEvents is not suitable (#240) for my use case.
However I noticed there's no `CreateKind` for kqueue implemented.
Is this a limitation of kqueue or an oversight?
If it's the former, is there maybe a workaround for this?

I can get a Remove(Any) event if the "Path" I'm watching is removed and now I need to know when a file is created at that "Path". As a workaround I tried to also watch the parent directory. This gives me an Access(Close(Write)) Event when a file is created. However, the path shown for this event is just the name of the watched parent directory, so I cannot map this event to the file I'm interested in.

@xanderio Ping

xanderio:
This PR fixes some shortcoming in the kqueue backend.

* The kqueue backend started as a copy of the inotify backend.
One major difference between kqueue and inotify was overlook by me,
inotify reports events for changes to files contained in a folder if the
folder is watched. Kqueue dosen't do this, each file has to be watch
individually.

* As kqueue dosen't provide this information directly we have to do some
trickery, that is keep a list of known file in a directory and compare
it to the directory after a Write event was emitted for that directory.

This fixes #365

### DIFF
--- a/src/kqueue.rs
+++ b/src/kqueue.rs
@@ -245,11 +245,8 @@ impl EventLoop {
             return self.add_single_watch(path, false);
         }
 
-        for entry in WalkDir::new(path)
-            .follow_links(true)
-            .into_iter()
-            .filter_map(filter_dir)
-        {
+        for entry in WalkDir::new(path).follow_links(true).into_iter() {
+            let entry = entry.map_err(map_walkdir_error)?;
             self.add_single_watch(entry.path().to_path_buf(), is_recursive)?;
         }
         self.kqueue.watch()?;
@@ -284,12 +281,8 @@ impl EventLoop {
                     .map_err(|e| Error::io(e).add_path(path.clone()))?;
 
                 if is_recursive || remove_recursive {
-                    for entry in WalkDir::new(path)
-                        .follow_links(true)
-                        .into_iter()
-                        .filter_map(filter_dir)
-                    {
-                        let p = entry.path().to_path_buf();
+                    for entry in WalkDir::new(path).follow_links(true).into_iter() {
+                        let p = entry.map_err(map_walkdir_error)?.path().to_path_buf();
                         self.kqueue
                             .remove_filename(&p, EventFilter::EVFILT_VNODE)
                             .map_err(|e| Error::io(e).add_path(p))?;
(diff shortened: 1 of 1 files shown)


src/kqueue.rs
```diff
@@ -156,12 +156,41 @@ impl EventLoop {
...
                             remove_watches.push(path.clone());
-                            Event::new(EventKind::Remove(RemoveKind::Any))
+                            Event::new(EventKind::Remove(RemoveKind::Any)).add_path(path)
                         }
 
-                        //data was written to this file
+                        // a write to a directory mean that a new file was created in it, let's
+                        // figure out which file this was
+                        kqueue::Vnode::Write if path.is_dir() => {
+                            // find which file is new in the directory be comparing it the our
```
0xpr03, line 165: ~~be~~by
it ~~the~~with our

src/kqueue.rs
```diff
@@ -156,12 +156,41 @@ impl EventLoop {
                         */
                         kqueue::Vnode::Delete => {
                             remove_watches.push(path.clone());
-                            Event::new(EventKind::Remove(RemoveKind::Any))
+                            Event::new(EventKind::Remove(RemoveKind::Any)).add_path(path)
                         }
 
-                        //data was written to this file
+                        // a write to a directory mean that a new file was created in it, let's
```
0xpr03, line 162: mean**s**

src/kqueue.rs
```diff
@@ -156,12 +156,41 @@ impl EventLoop {
...
+                        // figure out which file this was
+                        kqueue::Vnode::Write if path.is_dir() => {
+                            // find which file is new in the directory be comparing it the our
+                            // list of known watches
+                            if let Some(file) = std::fs::read_dir(&path)
+                                .unwrap()
```
0xpr03, line 168: please don't unwrap here, see #368

0xpr03 (changes requested):
Thanks for starting this. At best we'd add some kind of watcher restart feature, such that an invalid watcher state can be recovered from instead of rebuilding the whole watcher.

src/kqueue.rs
xanderio: I've update the code at this point so that we don't cause a panic, however i'm unware of any way to deal with this error. For now I just ignore it and return a Modify event for the directory. 

Question on the side: could we maybe add the log crate?

0xpr03: we can emit an error event like we do at other places via the channel

I can see the appeal of using logging, but it's also pretty invasive as we don't use this anywhere else and do have a back channel for emitting errors. Though it probably isn't used by many and is thus on par with logging things, the logging crate would require people to use a compatible logging crate to even see the events.

xanderio: I fixed this and update the commit

0xpr03 (approved):
looks good, thanks for investigating!

0xpr03:
Sadly the CI isn't happy https://github.com/notify-rs/notify/runs/4487903445?check_suite_focus=true
The error types don't seem to align.

xanderio:
That issue should be fixed now, totally missed it. A simple `cargo check` did reveal it, luckily there the CI :D

xanderio:
@0xpr03 is there anything preventing this from getting merged?

0xpr03:
Sorry it's just on me to find the time

0xpr03 (approved):

PR #373, opened 2021-12-07
inotify: Ignore any error sending the `RenameTimeout` event.

jorendorff:
The `unwrap()` call here causes thread panics in our application. We notice them because we use `set_panic_hook`. I think ordinarily a panic here would be printed to stderr and otherwise ignored.

For us, I think the usual pattern is that several files are renamed at once, and as soon as we notice the first rename, we drop the watcher. Therefore, when this thread wakes up, it's trying to send an event to a queue that has been closed.

### DIFF
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -416,9 +416,11 @@ impl EventLoop {
                             let cookie = rename_event.tracker().unwrap(); // unwrap is safe because rename_event is always set with some cookie
                             thread::spawn(move || {
                                 thread::sleep(Duration::from_millis(10)); // wait up to 10 ms for a subsequent event
-                                event_loop_tx
-                                    .send(EventLoopMsg::RenameTimeout(cookie))
-                                    .unwrap();
+
+                                // An error here means the other end of the channel was closed, a thing that can
+                                // happen normally.
+                                let _ = event_loop_tx
+                                    .send(EventLoopMsg::RenameTimeout(cookie));
                                 waker.wake().unwrap();
                             });
                         }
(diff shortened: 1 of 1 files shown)


0xpr03:
Hm, I think the proper fix is to shutdown the whole loop in that case.

0xpr03:
Though I wonder why you're the first to report this ? Maybe other people start it once and then only destroy it during shutdown ?

jorendorff:
> Hm, I think the proper fix is to shutdown the whole loop in that case.

Maybe I'm confused here. I was thinking this error can only happen if the event loop _has_ shut down.

This is a special, awkward bit of code that sleeps for 10 msec and then tries to send a message. All the other places where code sends a message to this queue, there's a comment that says

```
        // we expect the event loop to live and reply => unwraps must not panic
```

Here, we can't make that assumption; the way the code's written, there's an unavoidable race condition.

> Though I wonder why you're the first to report this ? Maybe other people start it once and then only destroy it during shutdown ?

Quite possible. But also, the panic always happens in a thread created by inotify, and the thread has no further responsibilities, so (unless you have `panic=abort` set) it doesn't hurt anything for that thread to panic. Maybe folks are just ignoring it.

0xpr03:
Yeah right I totally missed that thread creation and thought I'm inside the event loop.

0xpr03:
Though in that case you might want to also prevent a crash when the waker doesn't react any more.

jorendorff:
Ah, right, that makes sense. Added that.

0xpr03 (approved):
Thanks!

PR #377, opened 2022-01-03
Add more Debug/Copy trait impls

tmfink:
Also adds other useful impls.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,7 +3,7 @@
 use std::time::Duration;
 
 /// Indicates whether only the provided directory or its sub-directories as well should be watched
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
 pub enum RecursiveMode {
     /// Watch all sub-directories as well, including directories created after installing the watch
     Recursive,
@@ -24,7 +24,7 @@ impl RecursiveMode {
 /// Runtime configuration items for watchers.
 ///
 /// See the [`Watcher::configure`](../trait.Watcher.html#tymethod.configure) method for usage.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum Config {
     /// Enable or disable emitting precise event classification.
     ///
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -7,6 +7,7 @@ use super::event::*;
 use super::{Error, EventHandler, RecursiveMode, Result, Watcher};
 use filetime::FileTime;
 use std::collections::HashMap;
+use std::fmt::Debug;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::{
@@ -17,11 +18,13 @@ use std::thread;
 use std::time::{Duration, Instant};
 use walkdir::WalkDir;
 
+#[derive(Debug)]
 struct PathData {
     mtime: i64,
     last_check: Instant,
 }
 
+#[derive(Debug)]
 struct WatchData {
     is_recursive: bool,
     paths: HashMap<PathBuf, PathData>,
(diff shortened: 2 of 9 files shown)


src/config.rs
```diff
@@ -24,7 +24,7 @@ impl RecursiveMode {
 /// Runtime configuration items for watchers.
 ///
 /// See the [`Watcher::configure`](../trait.Watcher.html#tymethod.configure) method for usage.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```
0xpr03, line 27: Why do you need Hash ?

0xpr03 (changes requested):
I don't see why everything has to be Hash and Copy, adding more compile time.
Copy makes everything unable to integrate anything that is clone, so any changes that require clone will be API breaking. And specifically for watchers it is not advised to implement Clone/Copy.

Also please don't format stuff unrelated to your PR, this makes it much harder to review.

tmfink:
> I don't see why everything has to be Hash and Copy, adding more compile time.

By making your types implement common traits (that make sense), it's easier to use the library. Due to orphan rules, it would be impossible for a consumer of the library to put the type in a `HashMap` or `HashSet` (without using a newtype wrapper and inventing a hash function).  The [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits) API guideline has a good write-up.

Also, the difference compile-time is so small it's difficult to measure.

> Copy makes everything unable to integrate anything that is clone, so any changes that require clone will be API breaking (especially config is such a candidate).

I'll restrict the `Copy` impls to the "C-like" enums (no associated data).
Tell me what you think it the next round of review.  :wink: 

Also, it's already a breaking change to change variants of a public enum or to add variants to a non-`non_exhausitive` public enum (like `Config`).

> And specifically for watchers it is not advised to implement Clone/Copy.

I'll remove the watcher Clone/Copy.
It's nice to get review from someone who knows the codebase better. :smile:

> Also please don't format stuff unrelated to your PR, this makes it much harder to review.

Sure--I kept them in separate commits, but I'll drop them.
The [PR template](https://raw.githubusercontent.com/notify-rs/notify/adbbd338812a569dba3683e6cbdeb1f8f5976612/.github/PULL_REQUEST_TEMPLATE.md) is a little misleading:

    Running `cargo fmt` and/or `cargo clippy` is NOT required but appreciated!

src/config.rs
```diff
@@ -3,7 +3,7 @@
 use std::time::Duration;
 
 /// Indicates whether only the provided directory or its sub-directories as well should be watched
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
 pub enum RecursiveMode {
```
0xpr03, line 7: For this and Config I can't think of a reason to stuff them (indirectly) into a hashmap, but ok.

src/poll.rs
```diff
@@ -17,11 +18,13 @@ use std::thread;
 use std::time::{Duration, Instant};
 use walkdir::WalkDir;
 
+#[derive(Debug)]
 struct PathData {
     mtime: i64,
     last_check: Instant,
 }
 
+#[derive(Debug)]
```
0xpr03, line 27: This one isn't exposed, so Debug isn't necessary.

src/poll.rs
```diff
@@ -17,11 +18,13 @@ use std::thread;
 use std::time::{Duration, Instant};
 use walkdir::WalkDir;
 
+#[derive(Debug)]
```
0xpr03, line 21: This one isn't exposed, so Debug isn't necessary.

0xpr03 (changes requested):
I'd remove the debug traits on internal stuff. The changelog addition is nice, though we haven't updated that in a longer time. I'm still fetching all changes manually per release and the "overall" thing will have to be regenerated anyway ;)

0xpr03:
> non-non_exhausitive 

True, which is a reason we should probably make the fields non-pub and use a "builder" pattern.

> it's easier to use the library

Yes and no. Not adding copy to bigger types prevents you from accidentally copying stuff instead when you'd probably rather use a reference.

src/poll.rs
tmfink: This `Debug` impl is used by the `WatchData` impl

src/poll.rs
tmfink: This `Debug` impl (and `PathData`) is used by the `Debug` impl for `PollWatcher` (for the `.watches` field).

Would you prefer to leave the `.watches` field out of the Debug impl?

0xpr03: it's fine, it does expose some inner workings, but debug impl isn't stable per semver AFAIK, so reducing the printed info if relevant should be fine

0xpr03:
Thanks! When CI finished I'll merge it.

PR #378, opened 2022-01-07
Add Debug impl to fs/windows watchers

tmfink:
Additional Debug derives are required in order to implement Debug for
`FsEventWatcher` and `ReadDirectoryChangesWatcher`.

Also, print `PollWatcher` `.event_handler` field as a pointer.


JohnTitor (approved):
Seems fine to me

PR #383, opened 2022-02-01
Name all threads

Jasper-Bekkers:
It's a small change to give a name to all threads notify creates, this way it's easy to pinpoint where they come from in a debugger and profiler.

Edit: cargo fmt is a separate commit so it's a bit easier to see the changes.
Edit 2: `cargo build --workspace --all-targets` doesn't seem to build cleanly, and `cargo clippy --workspace --all-targets` also doesn't seem to be clean on main?
Edit 3: in our own codebase we've added this to clippy.toml which could be interesting to add here as well.

```
disallowed-methods = [
    # We require all of our threads to be named, use std::thread::Builder instead
    # to provide a name for the thread.
    "std::thread::spawn",
]
```

### DIFF
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -457,7 +457,7 @@ impl FsEventWatcher {
                     fs::FSEventStreamInvalidate(stream);
                     fs::FSEventStreamRelease(stream);
                 }
-            });
+            })?;
         // block until runloop has been sent
         self.runloop = Some((rl_rx.recv().unwrap().0, thread_handle));
 
--- a/src/fsevent.rs
+++ b/src/fsevent.rs
@@ -432,30 +432,32 @@ impl FsEventWatcher {
         // channel to pass runloop around
         let (rl_tx, rl_rx) = unbounded();
 
-        let thread_handle = thread::Builder::new().name("notify-rs fsevents".to_string()).spawn(move || {
-            let stream = stream.0;
+        let thread_handle = thread::Builder::new()
+            .name("notify-rs fsevents".to_string())
+            .spawn(move || {
+                let stream = stream.0;
 
-            unsafe {
-                let cur_runloop = cf::CFRunLoopGetCurrent();
+                unsafe {
+                    let cur_runloop = cf::CFRunLoopGetCurrent();
 
-                fs::FSEventStreamScheduleWithRunLoop(
-                    stream,
-                    cur_runloop,
-                    cf::kCFRunLoopDefaultMode,
-                );
-                fs::FSEventStreamStart(stream);
+                    fs::FSEventStreamScheduleWithRunLoop(
+                        stream,
+                        cur_runloop,
+                        cf::kCFRunLoopDefaultMode,
+                    );
+                    fs::FSEventStreamStart(stream);
 
-                // the calling to CFRunLoopRun will be terminated by CFRunLoopStop call in drop()
-                rl_tx
-                    .send(CFSendWrapper(cur_runloop))
-                    .expect("Unable to send runloop to watcher");
+                    // the calling to CFRunLoopRun will be terminated by CFRunLoopStop call in drop()
+                    rl_tx
+                        .send(CFSendWrapper(cur_runloop))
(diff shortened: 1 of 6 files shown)


Jasper-Bekkers:
Audit fails - but seems unrelated to my changes.

JohnTitor:
The idea itself makes sense to me! I'd like to hear @0xpr03's thoughts about event name bikeshedding.

Jasper-Bekkers:
Feel free to suggest different names, and I'll update it. I figured having a common prefix for them would be nice, so it's easy to see they're from this crate.

JohnTitor:
Seems they're out until Feb. 6th, wait for a while :)

src/poll.rs
```diff
@@ -76,106 +76,113 @@ impl PollWatcher {
...
-                                        Some(PathData {
-                                            mtime: old_mtime, ..
-                                        }) => {
-                                            if mtime > old_mtime {
-                                                let kind = MetadataKind::WriteTime;
-                                                let meta = ModifyKind::Metadata(kind);
-                                                let kind = EventKind::Modify(meta);
-                                                let ev = Event::new(kind).add_path(watch.clone());
-                                                event_handler(Ok(ev));
+                        for (
```
0xpr03, line 95: is all this indentation noise from rustfmt ?

src/poll.rs
```diff
@@ -76,106 +76,113 @@ impl PollWatcher {
...
-                                    let depth = if is_recursive { usize::max_value() } else { 1 };
-                                    for entry in WalkDir::new(watch)
-                                        .follow_links(true)
-                                        .max_depth(depth)
-                                        .into_iter()
-                                        .filter_map(|e| e.ok())
-                                    {
-                                        let path = entry.path();
+                                    } else {
+                                        let depth =
```
0xpr03, line 138: same here for the newline

tests/race-with-remove-dir.rs
```diff
@@ -10,14 +10,16 @@ fn test_race_with_remove_dir() {
 
     {
         let tmpdir = tmpdir.path().to_path_buf();
-        thread::spawn(move || {
-            let mut watcher = notify::recommended_watcher(move |result| {
-                eprintln!("received event: {:?}", result);
-            })
-            .unwrap();
+        let _ = thread::Builder::new()
+            .name("notify-rs test-race-with-remove-dir".to_string())
```
0xpr03, line 14: I'm not sure we want thread names in tests, as it'll just add more code noise, but I guess its fine

src/fsevent.rs
```diff
@@ -432,30 +432,32 @@ impl FsEventWatcher {
...
-                    .expect("Unable to send runloop to watcher");
-
-                cf::CFRunLoopRun();
-                fs::FSEventStreamStop(stream);
-                fs::FSEventStreamInvalidate(stream);
-                fs::FSEventStreamRelease(stream);
-            }
-        });
+        let thread_handle = thread::Builder::new()
+            .name("notify-rs fsevents".to_string())
```
0xpr03, line 436: maybe "notify-rs fsevent loop" ?

0xpr03 (commented):
Hey thanks for the PR.

I thought a little bit about the naming and I think adding something like "event loop" would be good, if we want to drop a hint for people debugging / inspecting the threads.

I marked two things because there is a lot of indentation noise I'd like to avoid if possible. If that's from rustfmt I guess we'll have to live with it.

src/poll.rs
Jasper-Bekkers: Yeah - it just jumps in a bit because the scope block indents one further.

src/fsevent.rs
Jasper-Bekkers: Done.

tests/race-with-remove-dir.rs
Jasper-Bekkers: I had the same doubts - but I figured why not.

0xpr03 (approved):
thanks

PR #395, opened 2022-03-16
Fix readme and examples

### LINKED ISSUE #393: `watcher` not available for import
#### System details

- OS/Platform name and version: MacOS 11.6.4
- Rust version (if building from source): `rustc --version`: `rustc 1.61.0-nightly (285fa7ecd 2022-03-14)`
- Notify version (or commit hash if building from git): `5.0.0-pre.14`

I'm trying to use the examples from the docs for version `5.0.0-pre.14` (or any v5 version), but I'm getting the following error:

```
error[E0432]: unresolved import `notify::watcher`
  --> src/lib.rs:14:57
   |
14 | use notify::{RecommendedWatcher, RecursiveMode, Result, watcher};
   |                                                         ^^^^^^^
   |                                                         |
   |                                                         no `watcher` in the root
   |                                                         help: a similar name exists in the module (notice the capitalization): `Watcher`

```

Not entirely clear what I'm doing wrong here, the code is taken directly from the docs/examples.

This can be replicated with the example direct from the docs:

```rs
use notify::{RecommendedWatcher, RecursiveMode, Result, watcher};
use std::time::Duration;

fn main() -> Result<()> {
    // Automatically select the best implementation for your platform.
    // You can also access each implementation directly e.g. INotifyWatcher.
    let mut watcher = watcher(Duration::from_secs(2))?;

    // Add a path to be watched. All files and directories at that path and
    // below will be monitored for changes.
    watcher.watch("/home/test/notify", RecursiveMode::Recursive)?;

    // This is a simple loop, but you may want to use more complex logic here,
    // for example to handle I/O.
    for event in &watcher {
        match event {
            Ok(event) => println!("changed: {:?}", event.path),
            Err(err) => println!("watch error: {:?}", err),
        };
    }

    Ok(())
}
```

### LINKED ISSUE #420: Remove Usage section from readme
It seems like the [with-a-channel](https://github.com/notify-rs/notify#with-a-channle ) section in the readme is not reflecting the newest API design where it says that there is a `Watcher::with_channel`, but there is no such API in the [docs](https://docs.rs/notify/5.0.0-pre.15/notify/index.html). I understand that the current 5.0.0 is not stablized yet, but I suggest to remove the entire Usage section in the README and point the usage to the API docs to reduce confusions.

0xpr03:
Add a smoke test for readme example, fixing the readme example itself. Also delete a lot of stuff that's not in v5 from the readme.

Move example files around so we actually implement example-workspaces correctly.

Fixes #393, fixes #420


README.md
```diff
@@ -36,110 +36,41 @@ crossbeam-channel = "0.4.0"
...
-let (tx, rx) = crossbeam_channel::unbounded();
-let mut watcher: RecommendedWatcher = Watcher::with_channel(tx, Duration::from_secs(2))?;
 
-for event in rx.iter() {
-    // ...
+    Ok(())
 }
 ```
 
-### With precise events
```
0xpr03, line 95: We do still talk about that in lib.rs. It's a planned feature but for now completely wrong. As we always emit precise events.

JohnTitor:
One possible way, to keep them in sync, is to use `#![doc = include_str!("../README.md")]` (the `extended_key_value_attributes` feature), which is stabilized in 1.54. I'm okay with or without it.

JohnTitor:
We don't have to do MSRV check on examples, change the below `if` conditions:
https://github.com/notify-rs/notify/blob/fcaeb0e982ab1fab9002e1eff0a37586f2391c9b/.github/workflows/main.yml#L45-L51

```diff
- if: matrix.version != 'stable' 
+ if: matrix.version == 'nightly'
```

(I don't remember why we ignore it on stable though...)

examples/watcher_kind/README.me
```diff
@@ -0,0 +1,3 @@
+# Watcher kind example
```
JohnTitor, line 1: Typo on file extension?

0xpr03: uh yeah, also thanks for reviewing but just to make sure: It's not finished, just had some time yesterday night

JohnTitor: Alright, ping me if it's ready :)

0xpr03:
Next step would be either a last RC or a release.

JohnTitor (approved):

JohnTitor:
> Next step would be either a last RC or a release.

I guess a RC release would be preferred as we have some changes since the last release?

PR #397, opened 2022-03-21
Add test for Debug impl for public types

tmfink:
Adds test that public types implement `Debug` trait

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,4 +301,35 @@ mod tests {
     fn test_object_safe() {
         let _watcher: &dyn Watcher = &NullWatcher;
     }
+
+    #[test]
+    fn test_debug_impl() {
+        macro_rules! assert_debug_impl {
+            ($t:ty) => {{
+                trait NeedsDebug: std::fmt::Debug {}
+                impl NeedsDebug for $t {}
+            }};
+        }
+
+        assert_debug_impl!(Config);
+        assert_debug_impl!(Error);
+        assert_debug_impl!(ErrorKind);
+        assert_debug_impl!(event::AccessKind);
+        assert_debug_impl!(event::AccessMode);
+        assert_debug_impl!(event::CreateKind);
+        assert_debug_impl!(event::DataChange);
+        assert_debug_impl!(event::EventAttributes);
+        assert_debug_impl!(event::Flag);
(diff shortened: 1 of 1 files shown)


tmfink:
We can ignore the `audit` test failure.

There is only one finding that has a solution:
```
Crate:         time
Version:       0.1.44
Title:         Potential segfault in the time crate
Date:          2020-11-18
ID:            RUSTSEC-2020-0071
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:      Upgrade to >=0.2.23
Dependency tree: 
time 0.1.44      
```

This finding only exists because:
1. `hot_reload_tide` example project depends on `tide 0.16.0` 
2. which has a transitive dependency on `chrono v0.4.19`
3. which has the non-default feature `oldtime` which provides compatibility with the old version of `time`

https://github.com/chronotope/chrono/blob/ea6412b3da34f216e3874b9d80dc95313118e894/README.md#duration

There's no security impact on users of this crate. Only the example project has the `time 0.1.44` dependency but does not enable the `chrono` feature that actually activates the dependency.

On a side note, because `cargo audit` output is so noisy, failures for the `audit` test should not cause all other tests to be cancelled ("fast fail"). I created #398 to handle.

0xpr03:
There is still a failure
https://github.com/notify-rs/notify/runs/5621135918?check_suite_focus=true#step:7:87
Which is why everything except bsd compiles.

0xpr03:
If you want to verify this for every watcher, you'd have to match the conditional compilation in lib.rs. Otherwise only the typical systems on linux/win/mac will be tested, but not fsevents on freebsd or such.

0xpr03:
But I'm also ok with leaving it like this.

tmfink:
> If you want to verify this for every watcher, you'd have to match the conditional compilation in lib.rs. Otherwise only the typical systems on linux/win/mac will be tested, but not fsevents on freebsd or such.

I purposely only included types that are defined for all platforms to avoid tedious repetition of the cfg attributes (which could easily diverge after refactoring). The important thing is that `RecommendedWatcher` is tested which is defined differently for each platform. As long as CI is run on these different platforms, then we should be (mostly) fine.

0xpr03:
Thanks, let's get this merged before it's getting more comments than LoC ;)

PR #398, opened 2022-03-21
ci: do not fail fast for `cargo audit`

tmfink:
Output of `cargo audit` can be very noisy, so don't cancel all other test cases if in the event of a failure. Maintainers can decide if the `cargo audit` failure merits an actual failure.

Based on test failures in #397.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,6 +77,8 @@ jobs:
 
   audit:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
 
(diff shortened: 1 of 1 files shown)


tmfink:
Looks like this change correctly allows the rest of the CI tests to run, even if `cargo audit` fails.

0xpr03:
#397 doesn't fail due to cargo audit. It's not neat that it always fails, but that's not the issue why your CI fails for that PR.

tmfink:
Agreed--closing.

PR #399, opened 2022-04-14
Fix build on FreeBSD

unrelentingtech:
Was failing due to non-exhaustive match which occurs because the kqueue crate has extra platform-specific enum variants.

While here, I noticed that the other BSDs had `mod kqueue` enabled but not the `pub use` re-export nor the `RecommendedWatcher`. Fixed that too.


JohnTitor (approved):
Makes sense to me 👍

0xpr03:
This may fix #396 thanks

0xpr03 (approved):

PR #402, opened 2022-04-30
Prepare 5.0.0-pre.15

0xpr03:


JohnTitor:
> The only thing I'm not so fond of is that without rustc 1.56 we can't simply enforce the new MSRV, which would probably make required upgrades easier to detect for downstream.

It's totally fine to bump up MSRV to 1.56 IMO, we could also switch to edition 2021. There's no need to happen it on this release but feel free to do so.

JohnTitor (approved):

0xpr03:
Ok, pushing to 1.56 actually breaks existing code due to now denied raw-pointer send/sync stuff.

PR #409, opened 2022-05-21
refactor: PollWatcher

visig9:
Refactor `PollWatcher` to make it clean for further develop.

This commit would not change any behaviors user observable (not include
`Debug` implementation) and public documentation.

---

I also add some `QUESTION` and `FIXME` in code comment to descript some weird (IMHO) behaviors. (Maybe we can fix it, or they have some pretty good reason so we should add description beside them)

If need any further change, please let me know, thanks.


visig9:
Found a different behavior: 

In `compare_contents` mode, original implementation use different `RandomState` (BuildHasher) instance in `watch()` and internal loop. So, event handler may immediate receive a `Modify(Data(Any))` event after user assign a new watch point.

See:

- https://github.com/notify-rs/notify/blob/0245625999e728ed3ce8dc6fdd5b6a7ad17e6510/src/poll.rs#L158-L162
- https://github.com/notify-rs/notify/blob/0245625999e728ed3ce8dc6fdd5b6a7ad17e6510/src/poll.rs#L311-L312

After refactor this behavior was gone. I think it's a bug so not worth to add it back, only note it here.

0xpr03 (approved):
Thanks! This looks very good.

Sometimes I'm not quite sure when cargo fmt does accept something or doesn't..

PR #413, opened 2022-06-07
Backport changes necessary for mio 0.8

upsuper:
This PR backports necessary changes for mio 0.8 support to notify 4.0 branch.

This is because mio 0.6 depends on miow 0.2.2 which has an open security issue: https://rustsec.org/advisories/RUSTSEC-2020-0080.html and thus it's triggering warnings for packages using notify 4.x.

Since 5.0 is still in pre-release, and seemingly has been in the status for a long time, it might be better backporting mio 0.8 to 4.x branch and release a new version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,9 @@ filetime = "0.2.5"
 walkdir = "^2.0.1"
 
 [target.'cfg(target_os="linux")'.dependencies]
+crossbeam-channel = "0.5.0"
 inotify = { version = "^0.7", default-features = false }
-mio = "^0.6.15"
-mio-extras = "^2.0.5"
+mio = { version = "0.8", features = ["os-ext"] }
 
 [target.'cfg(target_os="macos")'.dependencies]
 fsevent = "0.4"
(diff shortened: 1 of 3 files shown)


JohnTitor:
I'm hesitant to introduce a breaking change. And I'm currently considering making 5.0 release in the next few months and I'd go for it.

upsuper:
Oh yep, I mean 4.0, sorry. I don't think this is a breaking change as I'm not changing any API of this package as far as I can see?

JohnTitor:
@0xpr03 This would bump our MSRV up to 1.46 (https://github.com/tokio-rs/mio/blob/master/CHANGELOG.md#080) It makes some sense to resolve the security advisory issue and 1.46 is old enough, so I think it's an acceptable breakage. Thoughts?

JohnTitor:
Well, I just realized ubuntu-latest's failure was only on MSRV check, I shouldn't work on serious things tonight...

0xpr03:
> to resolve the security advisory issue and 1.46 is old enough

In itself yeah.
But as @xfix said, there is a backport for mio: https://github.com/yoshuawuyts/miow/pull/40

upsuper:
Oh, wasn't aware that it has a fix backported. We probably don't need this then.

0xpr03:
Still thank your for taking the time.

PR #417, opened 2022-06-16
Clean-up CI config

JohnTitor:


src/poll.rs
```diff
@@ -523,7 +525,7 @@ impl PollWatcher {
             .unwrap()
             .remove(path)
             .map(|_| ())
-            .ok_or_else(|| crate::Error::watch_not_found())
+            .ok_or_else(crate::Error::watch_not_found)
```
0xpr03, line 528: hm yeah should've watched out for that, probably let clippy do its thing

.github/workflows/main.yml
```diff
@@ -61,7 +72,6 @@ jobs:
         target:
           - x86_64-unknown-netbsd
           - x86_64-unknown-freebsd
-          - x86_64-apple-darwin
```
JohnTitor, line 64: @0xpr03 Removed x86_64-apple-darwin as we've already tested code with macOS on the above `ci` workflow. I think it should be fine but would like to mention it, just in case.

0xpr03:
Thanks, that'll make the CI usable again for other PRs

PR #419, opened 2022-06-21
poll-watcher: let all poisoned mutex just panic

visig9:
in poll-watcher, some place mutex poisoned cause panic, some other place just ignored. This PR try to fix this inconsistency by panic them all.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -461,15 +461,15 @@ impl PollWatcher {
                         break;
                     }
 
-                    // HINT: Make sure always lock in the same order to avoid deadlock.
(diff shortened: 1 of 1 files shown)


0xpr03:
I wouldn't panic them, I planned ton throwing errors for them

0xpr03:
But for know this might be an improvement over ignoring it..

0xpr03:
> Sorry to bother

No no, that's totally fine. And for everything you can't send the user (mutex that contains the sendback channel) it's valid that we don't ignore it..

PR #425, opened 2022-08-08
make crossbeam-channels optional

### LINKED ISSUE #380: Crossbeam breaks tokio::spawn
#### System details

- OS/Platform name and version: Linux 5.10.90
- Rust version (if building from source): 1.57
- Notify version (or commit hash if building from git): 15.0.0-pre3

Using a bunch of `tokio::spawn(async move` calls, i'm seeing the last in the series never complete. Upon digging into the GH issues for Tokio i found [this little gem](https://github.com/tokio-rs/tokio/issues/1388) which seems to say that crossbeam breaks tokio in exactly the way i'm seeing it break. The thing is, i'm not even using crossbeam in my code, i'm using tokio's unbounded channels but the presence of crossbeam is breaking things. When i make the tokio channel bounded it breaks completely - no `recv()` occurs. Removing notify from the codebase fixes the problem but its a critical piece of the tool being written.
Specifically when sending 4 items into a channel of `PathBufs` i see all 4 being `recv()`d  (i print them out at that point) with the result of the `recv` being passed to a `tokio::spawn(async move` which then prints the `PathBuf` again. However, only 3 of those 4 actually print from inside the spawned thread - the last one (always, it seems) never fires - in line with the bug referenced above.

#### What you expected

Expect every Tokio spawned task to fire all the time

#### What happened

Some compilations produce a working binary, and others, with no source changes but a `cargo clean` produce binaries which omit the last execution of the loop running off the tokio::mspc channel's receiver (the data is recv'd but the `spawn` call goes off into oblivion and never prints or completes). Sometimes using the variables inside the thread make it fire (like printing them a bunch of times), but something is breaking the scheduler (most likely crossbeam).

0xpr03:
Allows disabling crossbeam-channel by disabling default features.

This fixes #380

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,21 +37,31 @@ jobs:
           rustup toolchain install ${{ matrix.version }} --no-self-update
           rustup override set ${{ matrix.version }}
 
-      - name: check build
+      - name: check build serde,macos_kqueue for examples
         if: matrix.version != '1.56.0' && matrix.os == 'macos-latest'
-        run: cargo check --features=serde,macos_kqueue --examples
+        run: cargo check -p notify --features=serde,macos_kqueue --examples
 
-      - name: check build
+      - name: check build serde,macos_kqueue
         if: matrix.version == '1.56.0' && matrix.os == 'macos-latest'
-        run: cargo check --features=serde,macos_kqueue
+        run: cargo check -p notify --features=serde,macos_kqueue
 
-      - name: check build
+      - name: check build serde for examples
         if: matrix.version != '1.56.0' && matrix.os != 'macos-latest'
-        run: cargo check --features=serde --examples
+        run: cargo check -p notify --features=serde --examples
 
-      - name: check build
+      - name: check build serde
         if: matrix.version == '1.56.0' && matrix.os != 'macos-latest'
         run: cargo check --features=serde
+
+      - name: check build without crossbeam/default features
+        if: matrix.version == 'stable'
+        run: cargo check -p notify --no-default-features --features=macos_fsevent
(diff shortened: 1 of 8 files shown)


0xpr03:
I'll leave an updated documentation to #395

0xpr03:
One downside of this approach is that users will have to use
```toml
notify = { version = "5", default-features = false, feature=["macos_kqueue"] }
```

JohnTitor (approved):
Looks like a good approach!

> One downside of this approach is that users will have to use
> 
> ```toml
> notify = { version = "5", default-features = false, feature=["macos_kqueue"] }
> ```

I think it's acceptable as only those who want to disable it have to touch the file, that is existing users that don't have the crossbeam issue don't have to tweak their Cargo.toml.

nicoburns:
I wonder whether it would be worth switching wholesale to something like https://docs.rs/flume/latest/flume/ ? I believe it has very similar performance to crossbeam_channel. It might cause the same problem of course.

0xpr03:
Unbounded MPSC seems to be the only case crossbeam-channel doesn't win over flume according to their benchmarks. Replacing our mpsc stuff with spsc would probably make more sense.

As far as I know parts of crossbeam are considered for moving into std, which would give it more credibility.

One major reason for using corssbeam-channel is probably the thread parking when waiting for channel events.

0xpr03:
If flume does prevent tokio errors, I'd be willing to accept a PR. For now my main concern is resolving the tokio scheduler issues.

This PR makes it very easy to switch it out, so if any one wants to test this, feel free. As far as I know, you can't instantly trigger this scheduler hang, so help is welcome for this.

Issue #249, opened 2020-06-10
Release 5.0!


tokahuke:
Release candidate 2 is already 5 months in the air. Isn't is time to release it as "final" already?

JohnTitor:
Before releasing, I'd like to resolve #248 and audit test cases as 2668c204525fc081bd6001acb1e8797ae520a43c removed most of them.

pksunkara:
What happened to debounced watcher in the pre releases? Is it being removed for 5.0?

0xpr03:
It was removed due to problems, so we would have to re-implement it for a real 5.0 release.

pksunkara:
Ah, thanks for the answer. I would love to contribute since I kinda need it, but I am not knowledgable enough in this area. 😞

dragonnn:
@0xpr03 maybe using https://crates.io/crates/rjdebounce ? I used it outside notify in 5.0.0-pre3 and it works fine.
I need 5.0.0 because I need to have an option to do things without using thread channels

0xpr03:
> @0xpr03 maybe using https://crates.io/crates/rjdebounce ?

I don't think we want to add more dependencies just for debouncing. Also looking at the code this does not do what we want. The job of a debouncer it to record all events and only return you the ones that are the most important (create after write+delete) per file (notice directory deletions overriding subfiles..) after a timeout (or even without a timeout if something always receives an event => permanent write). Also we want to introduce the option of receiving all events and also the debounced version at the same time. Additionally we can't simply not run our edge code for X seconds as we have to catch all system events to handle recursive watching newely created files.

0xpr03:
> I need to have an option to do things without using thread channels

What exactly do you mean without threads ? Notify certainly doesn't support no-std and even without any debouncer we're [using](https://github.com/notify-rs/notify/blob/e69d5a23384dfb4d0ac2bc497e2658d327108177/src/inotify.rs#L127) a thread underneath.

xanderio:
After #335 and #337 have been merged, cloud we get a new prerelease?

lulujun2233:
Is there deadline or timeline to release 5.0 (with debouncer)?

jyn514:
I would find a release useful even without the debouncer, it would let me get rid of a duplicate `mio` dependency. Tokio has depended on mio 0.7 for a while now so any crate with both tokio and notify pulls in a duplicate mio dependency.

0xpr03:
I don't think you want me to release v5 in the current state.
[...]

PR #426, opened 2022-08-10
replace old config and allow initialization configuration

### LINKED ISSUE #418: Consider use dedicated configuration interface for every single watcher implementation.
I believe configurations should not be given in general level. Ref: https://docs.rs/notify/5.0.0-pre.15/notify/enum.Config.html

- `NoticeEvents`: Applicable to debounced watchers only.
- `OngoingEvents`: Applicable to debounced watchers only.
- `PreciseEvents`: The exactly meaning depend on watcher implemention detail. That mean user must know which watcher they want to used.

Further more, current interface also force API user need to understand which config option should be set in which location. For example if I want to use `PollWatcher`, I need to check following configuration location:

- Set `PreciseEvents` here: https://docs.rs/notify/5.0.0-pre.15/notify/trait.Watcher.html#method.configure
- Set `poll_interval` & `compare_contents` in here: https://docs.rs/notify/5.0.0-pre.15/notify/poll/struct.PollWatcher.html#method.with_config
- Set `RecursiveMode` in here: https://docs.rs/notify/5.0.0-pre.15/notify/trait.Watcher.html#tymethod.watch
    - I know this is watch point level config so can not setup at watcher level, but still a config and not easy to set. Suggestion: how about just split `.watch(path, recu_mode)` into 2 function like `.watch_exact(path)`, `.watch_recu(path)`.

`notify` can give every single watcher implementation a dedicated builder to let API user easy to config everything what they want. A drawback of builder pattern is, user can't change configurations after Watcher already started (If `notify` really want to allow user to do it)

0xpr03:
Removes all event-level configuration, which is completely unimplemented by all backends.
Replaces them with the init-critical PollWatcher config. Allows configuration during initialization. Uses a builder pattern with defaults to allow future settings.

This way you can setup a recommended watcher while still specifying the wanted PollWatcher settings and the debouncer can initiate a PollWatcher with `compare_contents`, something that can only be done pre runtime.

Related #249 
Closes #418

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,43 +25,6 @@ As used by: [alacritty], [cargo watch], [cobalt], [docket], [mdBook], [pax],
 [rdiff], [rust-analyzer], [timetrack], [watchexec], [xi-editor], [watchfiles],
 and others.
 
-## Base Installation
-
-```toml
-[dependencies]
-notify = "5.0.0-pre.15"
-```
-
-## Usage
-
-A basic example 
-
-```rust
-use notify::{RecommendedWatcher, RecursiveMode, Result, watcher};
-use std::time::Duration;
-
-fn main() -> Result<()> {
-    // Automatically select the best implementation for your platform.
-    // You can also access each implementation directly e.g. INotifyWatcher.
-    let mut watcher = watcher(Duration::from_secs(2))?;
-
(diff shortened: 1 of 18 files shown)


JohnTitor (approved):
LGTM 👍

PR #427, opened 2022-08-12
prepare 5.0.0-pre.16

0xpr03:
If merged, I'll publish the new debouncer and notify pre.16.


0xpr03:
This also removes the old v5 changelog, as I think it's mostly out of date, already happened or rather something we should put inside the upgrade doc.

The actual v5 changelog will be quite big..

JohnTitor (approved):

PR #429, opened 2022-08-14
fixup optional crossbeam feature selection in debouncer-mini

### LINKED ISSUE #433: Reminder: the debouncer lib.rs still has debug println
https://github.com/notify-rs/notify/blob/94f1680713036cd4744df4543f660a44a435be85/notify-debouncer-mini/src/lib.rs#L177
https://github.com/notify-rs/notify/blob/94f1680713036cd4744df4543f660a44a435be85/notify-debouncer-mini/src/lib.rs#L180
https://github.com/notify-rs/notify/blob/94f1680713036cd4744df4543f660a44a435be85/notify-debouncer-mini/src/lib.rs#L208

0xpr03:
Actually pass through the optional crossbeam dep from debouncer-mini to notify.

Also verify this in CI. (Requires a manual -p call to work.)

Also closes #433

### DIFF
--- a/notify-debouncer-mini/Cargo.toml
+++ b/notify-debouncer-mini/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "notify-debouncer-mini"
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 rust-version = "1.56"
 description = "notify mini debouncer for events"
@@ -19,7 +19,9 @@ name = "notify_debouncer_mini"
 path = "src/lib.rs"
 
 [features]
-default = ["crossbeam-channel"]
+default = ["crossbeam"]
+# can't use dep:crossbeam-channel and feature name crossbeam-channel below rust 1.60
+crossbeam = ["crossbeam-channel","notify/crossbeam-channel"]
 
 [dependencies]
 notify = "5.0.0-pre.16"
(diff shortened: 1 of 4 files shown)


0xpr03:
The bad part: if we want to name the feature `crossbeam-channel`, as used in notify, we'll have to raise the MSRV to 1.60.
So this is my workaround.

0xpr03:
CC @JohnTitor for pro/contra

notify-debouncer-mini/Cargo.toml
```diff
@@ -19,7 +19,9 @@ name = "notify_debouncer_mini"
 path = "src/lib.rs"
 
 [features]
-default = ["crossbeam-channel"]
+default = ["crossbeam"]
+# can't use dep:crossbeam-channel and feature name crossbeam-channel below rust 1.60
+crossbeam = ["crossbeam-channel","notify/crossbeam-channel"]
```
JohnTitor, line 24: Another way would be renaming our dependency like:

```toml
[dependencies]
crossbeam = { package = "crossbeam-channel", version = "..." }
```

JohnTitor: But the `crossbeam` feature name doesn't sound that bad, so I'm fine either way :)

0xpr03: oh didn't think of renaming them, but yeah I'll rename the feature in the debouncer

PR #431, opened 2022-08-14
improve upgrade docs and fix Config link

0xpr03:


0xpr03:
cc #249 maybe people over there find additional things to add for a v4 to v5 transition

PR #432, opened 2022-08-15
update minimum walkdir version to 2.2.2

trinity-1686a:
While trying out `notify 5.0.0-pre.16`, I got some issues as it declare a dependency on `walkdir 2.0.1`, but uses a [function added in `2.2.2`](https://docs.rs/walkdir/2.2.2/walkdir/struct.DirEntry.html#method.into_path). This patch aims to fix that little issue.

I've tested the RC for my (arguably simple) use case, and on Linux it seems to work fine. I'll test on MacOS and Windows in the following days, and report if I got any issue.


trinity-1686a:
CI seems to have failed due to "1 cancelled checks" for macos-latest nightly. I'm not sure what that means or whether I can do something about it

JohnTitor (approved):
Thanks!

Issue #434, opened 2022-08-17
Rescan flag is correctness hazard


ijackson:
Almost every caller of `notify` needs to handle rescan events, to be correct.

However, in order to discover that this thing even exists it is necessary to dig many levels deep in the documentation.  In particular, starting from `Event`, one must navigate the following:

 * `attrs: EventAttributes`

    Additional attributes of the event.

    Arbitrary data may be added to this field, without restriction beyond the Sync and Clone properties. Some data added here is considered for comparing and hashing, but not all: at this writing this is Tracker, Flag, Info, and Source.

This does not look like somethilng one needs to consider in a simple program.  Most readers will only look here if they are trying to find how to do something complicated and can't find the API for it elsewhere.

 * `pub fn flag(&self) -> Option<Flag>`

    Retrieves the Notify flag for an event directly, if present.

Again, this text does not even hint that looking at this flag is necessary for a reliable program.

 * Enum `notify::event::Flag`

    Special Notify flag on the event.

   This attribute is used to flag certain kinds of events that Notify either marks or generates in particular ways.

Once again, the reader is discouraged from reading on.  Eventually we come to this:

 * Variants: `Rescan`

    Rescan notices are emitted by some platforms (and may also be emitted by Notify itself). They indicate either a lapse in the events or a change in the filesystem such that events received so far can no longer be relied on to represent the state of the filesystem now.

   An application that simply reacts to file changes may not care about this. An application that keeps an in-memory representation of the filesystem will need to care, and will need to refresh that representation directly from the filesystem.

Here, at least, the careful reader will notice that actually this flag means "we have missed stuff".  But you have to read this text quite carefully.  The critical information is in the last sentence.  (I don't really agree that  "An application that simply reacts to file changes may not care about this".  Any application that doesn't also separately poll will malfunction if rescan lapses occur but or not handled.)

I think this API is setting up developers for failure.  Specifically, it is setting up developers for writing programs which mysteriously lose file updates but only when the system is under load.

IMO `Event` should have a structure which makes it impossible to access `paths` or `kind` without checking for `Rescan`.  After all, a `Rescan` event conceptually applies to all paths and all kinds, so an application which filters on paths or kind must know about `Rescan`.

0xpr03:
(I took the liberty of formatting your comment to better distinguish the quotes)

0xpr03:
If possible we could move that Event towards its own EventKind

And think about how we handle this for the debouncer (re-initialize?)

[...]

PR #435, opened 2022-08-17
add accessor for whether rescan is required on Event

trinity-1686a:
this is a low-effort/no-breakage solution to #434


0xpr03:
thanks!

PR #440, opened 2022-08-31
debouncer: correctly document crossbeam feature

0xpr03:
Marks 0.2.1 to fixup feature name in docs

### DIFF
--- a/notify-debouncer-mini/src/lib.rs
+++ b/notify-debouncer-mini/src/lib.rs
@@ -42,10 +42,11 @@
 //!
 //! # Features
 //!
-//! The following feature can be turned on or off.
+//! The following crate features can be turned on or off in your cargo dependency config:
 //!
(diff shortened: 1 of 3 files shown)


notify-debouncer-mini/src/lib.rs
```diff
@@ -44,7 +44,8 @@
 //!
 //! The following feature can be turned on or off.
 //!
-//! - `crossbeam-channel` enabled by default, adds DebounceEventHandler support for crossbeam channels.
+//! - `crossbeam` enabled by default, adds DebounceEventHandler support for crossbeam channels.
+//!   Also enables crossbeam-channel in the re-exported notify. You my want to disable this when using the tokio async runtime.
```
JohnTitor, line 48:
```suggestion
//!   Also enables crossbeam-channel in the re-exported notify. You may want to disable this when using the tokio async runtime.
```

Issue #423, opened 2022-07-20
`Function not implemented (os error 38)` when running on docker in M1 MacOS 


samuelcolvin:
#### System details

See https://github.com/samuelcolvin/watchfiles/issues/167 for initial report

- OS/Platform name and version: docker running on M1 mac with the `--platform linux/amd64` flag set
- Rust version (if building from source): `rustc --version`: `rustc 1.61.0 (fe5b13d68 2022-05-18)`
- Notify version (or commit hash if building from git): `=5.0.0-pre.15`
- If you're coming from a project that makes use of Notify, what it is, and a link to the downstream issue if there is one:
  - [watchfiles](https://github.com/samuelcolvin/watchfiles), issue: https://github.com/samuelcolvin/watchfiles/issues/167
  - and upstream from there: https://github.com/adamchainz/django-watchfiles/issues/3
- Filesystem type and options: Mac OS with M1 architecture running docker using the `python:3.10` image

#### What you did (as detailed as you can)

When running watchfiles, the following error occurs: `Function not implemented (os error 38)`

In watchfiles code, the error is caught and raised [here](https://github.com/samuelcolvin/watchfiles/blob/93f744dff65c5474da6f60b8b582c8dc37f6f91a/src/lib.rs#L124).

This occurs when using `RecommendedWatcher`, when using `PollWatcher` everything seems to be fine.

There are a number of other similar reports of the same error message on other rust packages running in docker in the same way, see https://github.com/adamchainz/django-watchfiles/issues/3. Those seem to related to missing binaries for MacOS M1, but watchfiles provides (AFAIK) all applicable binaries.

I think this is a problem with docker not providing the underlying functions which notify expects to find.

I _think_ I can catch this error and fall back to `PollWatcher` in watchfiles but I wanted to report the issue here in case you had any insights or suggestions.

samuelcolvin:
Ye, I get the problem.

You won't be able to test on gh-actions or any other cloud service AFAIK.

I don't know of any to actively detect, but I guess there must be a way?

messense:
> Is there a specific conditional compilation [architecture](https://doc.rust-lang.org/reference/conditional-compilation.html?search=#target_arch) we could detect for this ?

I don't think there is. The target architecture is x86_64. You'd need to detect QEMU emulation in runtime, but that's not pretty either. https://stackoverflow.com/a/71643762

0xpr03:
So the question is: Do we want to add this error-code workaround into notify ?

[...]

PR #441, opened 2022-08-31
Add method for native watcher initialization with fallback

0xpr03:


0xpr03:
This is a draft because:
1. I'd like to also change the `recommended_watcher` function but would have to change its result value. Not sure if we can do that semver compatible (or just ignore it).
2. I've opted for an enum type instead of Box<dyn Watcher>, I could also imagine using that.
3. I've only handled the specific #423 error. We could also fallback for any error.
4. Do we want to report the original error with the fallback ? Add some additional error kind "unavailable API" to at least report it?

notify/src/lib.rs
```diff
@@ -380,6 +380,58 @@ where
     RecommendedWatcher::new(event_handler, Config::default())
 }
 
+/// Method for creating the `RecommendedWatcher` for the current platform
+/// and falling back if the recommended API is not available.
+/// 
+/// Example use case is [#423] where Docker on M1 in qemu does not expose the inotify OS API.
```
JohnTitor, line 386: I guess the link `[#423]` doesn't work.

JohnTitor:
>     1. I'd like to also change the `recommended_watcher` function to just call the new one, giving it more value, but would have to change its result value. Not sure if we can do that semver compatible (or just ignore it).

I'd call it a breaking change if it changes behavior.

>     2. I've opted for an enum type instead of `Box<dyn Watcher>`, I could also imagine using that.

I think it's totally fine while we have `take_boxed`.

>     3. I've only handled the specific [`Function not implemented (os error 38)` when running on docker in M1 MacOS  #423](https://github.com/notify-rs/notify/issues/423) error. We could also fallback for any error.

These could come later :)

>     4. Do we want to report the original error with the fallback ? Add some additional error kind "unavailable API" to at least report it?

I don't have a strong opinion on that, maybe we could prepare an optional value?

>     5. This requires `EventHandler` to be declared either `Copy` or `Clone` for this fallback function. Even more of a SemVer break for `recommended_watcher`

Hmm, that sounds like a rather difficult issue and I cannot prepare a solution for that soon (I'm now fixing my dev env and cannot check your code deeply).

notify/src/lib.rs
```diff
@@ -380,6 +380,58 @@ where
...
+{
+    // All recommended watchers currently implement `new`, so just call that.
+    match RecommendedWatcher::new(event_handler, config.clone()) {
+        Ok(v) => Ok(WatcherFallback::Native(v)),
+        Err(e) => {
+            match &e.kind {
+                #[allow(unused)]
+                ErrorKind::Io(io_error) => {
+                    #[cfg(target_os = "linux")]
+                    if io_error.raw_os_error() == Some(38) {
```
samuelcolvin, line 401: Is there a documented meaning to 38 or indeed other codes somewhere? I suspect we should fallback in more cases but I have no idea how to find out which ones should be an error and which should fallback.

notify/src/lib.rs
```diff
@@ -380,6 +380,58 @@ where
...
+    Native(RecommendedWatcher),
+    /// Fallback watcher, known platform issue occured
+    /// 
+    /// For example the Docker with Linux on MacOS M1 bug
+    Fallback(PollWatcher)
+}
+
+impl WatcherFallback {
+    /// Returns the watcher inside
+    pub fn take_boxed(self) -> Box<dyn Watcher>{
```
samuelcolvin, line 427: personally I would prefer that this implemented `watch`. Is that very verbose/difficult?

samuelcolvin (commented):
broadly seems fine to me, I'll probably stick with that I have now in watchfiles though since it's working well.

notify/src/lib.rs
0xpr03: that's a very good idea! and it would make the enum more worthwhile (because otherwise I don't actually see much of a value over directly boxing it, except for claiming the allocation-free throne, while not being useful at all)

0xpr03: Ok I tried this.. And it gets really awkward.

- We can't implement `Watcher::new` due to missing `Copy`/`Clone `on `event_handler`.
- We can't implement a real `kind()`, as we don't have access to `self`. This kind of API is specifically not designed for runtime-selected watchers it seems. We could still implement the functions without the trait, or leave them `unimplemented!()` for the new and the additional watcher-kind as in my new code.

See also the added comments.

notify/src/lib.rs
0xpr03: fixed

0xpr03:
Please review again, the addition of Deref/DerefMut allows direct access to the watcher in our enum, making the whole stack allocation worthwhile, such that boxing is not required. [Example](https://github.com/notify-rs/notify/blob/31d9860912cf261d8fa69413e004ef4ea8ea253d/examples/multiple_backends.rs#L27-L31)

Open Questions:
- Do we deprecate `recommended_watcher()` to reflect that it's kind of unhelpful as a function (in light of RecommendedWatcher) ?
- Do we migrate all examples (except for one) and the docs to recommend using `recommended_watcher_fallback` over `RecommendedWatcher` ? Technically it is the only way which we can get to actually work on all systems (and can work around future issues). In a way comptime-selection was fun while it lasted. On the other hand a lot of people don't seem to care about the specific issue on M1 + Docker (you can also say people should not expect such weird emulator setups to work), but I don't see a drawback for people already using `RecommendedWatcher` to migrate to the dynamic approach.
- Bikeshed: Is there a better name than `recommended_watcher_fallback` ? It's kinda long when for normal users I'd say it is the recommended way.
- I have no good answer to the copy/clone requirement of the fallback logic.

notify/src/lib.rs
0xpr03: Hm, here is [the manpage](https://man7.org/linux/man-pages/man3/errno.3.html) though I prefer [this list](https://www.thegeekstuff.com/2010/10/linux-error-codes/)
Problem is that this does not tell us which of these could be emitted by our specific syscall.

juliusl:
> Bikeshed: Is there a better name than recommended_watcher_fallback ? It's kinda long when for normal users I'd say it is the recommended way. @0xpr03 

How about `ensure_watcher`? From my experience, ensure has a stronger implication that a resource will be obtained.

Anutrix:
Why was this closed?

PR #454, opened 2022-11-16
kqueue: Only make a single syscall to add all the watches

alexkirsz:
Hey!

We use notify v4 in [`turbopack`](https://github.com/vercel/turbo) and we've recently tried updating to v5 to use the `kqueue` implementation, as we found it to be much faster than FSEvents at detecting updates.

However, we also noticed a very large drop in performance when watching large directories. After a bit of investigating, it turns out that `notify` makes a `kevent` syscall for each file it watches. Watching a folder with 15k files ends up taking about 11.5s.

This PR changes this behavior to ensure we only make a single `kevent` call when adding multiple files at the same time. This improves the above case to take 900ms.

I've also filed [an issue with `kqueue`](https://gitlab.com/rust-kqueue/rust-kqueue/-/issues/12) to fix another performance problem, which makes watching n files O(n²). This would further improve to 400ms.

Finally, while profiling the above issue, I noticed that during the 12s of watching, only 2 cores of my computer were working. I have a prototype version that uses multiple threads to open files, which might be faster when watching large directories, but it would also require changes to `kqueue`, so I'm filing it away for now.

### DIFF
--- a/notify/src/kqueue.rs
+++ b/notify/src/kqueue.rs
@@ -288,13 +288,15 @@ impl EventLoop {
         // If the watch is not recursive, or if we determine (by stat'ing the path to get its
         // metadata) that the watched path is not a directory, add a single path watch.
         if !is_recursive || !metadata(&path).map_err(Error::io)?.is_dir() {
-            return self.add_single_watch(path, false);
+            self.add_single_watch(path, false)?;
+        } else {
+            for entry in WalkDir::new(path).follow_links(true).into_iter() {
+                let entry = entry.map_err(map_walkdir_error)?;
+                self.add_single_watch(entry.path().to_path_buf(), is_recursive)?;
+            }
         }
 
-        for entry in WalkDir::new(path).follow_links(true).into_iter() {
-            let entry = entry.map_err(map_walkdir_error)?;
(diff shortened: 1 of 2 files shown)


0xpr03:
cc @xanderio who added this initially for BSD

xanderio (approved):

notify/src/kqueue.rs
```diff
@@ -314,7 +316,7 @@ impl EventLoop {
             .add_filename(&path, event_filter, filter_flags)
             .map_err(|e| Error::io(e).add_path(path.clone()))?;
         self.watches.insert(path, is_recursive);
-        self.kqueue.watch()?;
```
0xpr03, line 317: Can we add a comment to `add_single_watch` that it requires calling `kqueue.watch` by the caller ?

0xpr03:
I'm intending to accept this. I'd like to have some confirmation from BSD & Apple folks that this actually works as intended for single files & recursive folders.

notify/src/kqueue.rs
alexkirsz: Done in https://github.com/notify-rs/notify/pull/454/commits/f6da0db4d5bb236b7f4f71e093ec21aa6cb0f0ee

PR #457, opened 2022-12-05
Use official windows-sys crate for WinAPI instead of deprecated winapi crate

GamePad64:
`winapi` crate is kinda deprecated. `windows` and `windows-sys` can be used instead. They are maintained directly by Microsoft and are actively developed.
`windows-sys` is a trimmed down version of `windows`. It offers better compilation time, and it is released once per 6 months.

I removed winapi dependency and replaced in with windows-sys.

### DIFF
--- a/notify/src/windows.rs
+++ b/notify/src/windows.rs
@@ -5,17 +5,7 @@
 //!
 //! [ref]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa363950(v=vs.85).aspx
 
-use winapi::shared::minwindef::TRUE;
-use winapi::shared::winerror::ERROR_OPERATION_ABORTED;
-use winapi::um::fileapi;
-use winapi::um::handleapi::{self, INVALID_HANDLE_VALUE};
-use winapi::um::ioapiset;
-use winapi::um::minwinbase::{LPOVERLAPPED, OVERLAPPED};
-use winapi::um::synchapi;
-use winapi::um::winbase::{self, INFINITE, WAIT_OBJECT_0};
-use winapi::um::winnt::{self, FILE_NOTIFY_INFORMATION, HANDLE};
-
-use crate::{bounded, unbounded, BoundSender, Receiver, Sender, Config};
+use crate::{bounded, unbounded, BoundSender, Config, Receiver, Sender};
 use crate::{event::*, WatcherKind};
 use crate::{Error, EventHandler, RecursiveMode, Result, Watcher};
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


notify/src/windows.rs
```diff
@@ -29,6 +19,23 @@ use std::ptr;
 use std::slice;
 use std::sync::{Arc, Mutex};
 use std::thread;
+use windows_sys::Win32::Foundation::{
+    CloseHandle, ERROR_OPERATION_ABORTED, HANDLE, INVALID_HANDLE_VALUE, WAIT_OBJECT_0,
+};
+use windows_sys::Win32::Storage::FileSystem::{
+    CreateFileW, ReadDirectoryChangesW, FILE_ACTION_ADDED, FILE_ACTION_MODIFIED,
```
0xpr03, line 26: I think it's better to just do `::*` for this case.

0xpr03 (approved):
thanks!

notify/src/windows.rs
GamePad64: windows_sys::Win32::Storage::FileSystem is pretty big, so `::*` may lead to great scope pollution…
I am thinking about `use windows_sys::Win32::Storage::FileSystem as fileapi`, so it will have a short name, like in winapi.

notify/src/windows.rs
```diff
@@ -415,9 +420,8 @@ impl ReadDirectoryChangesWatcher {
     ) -> Result<ReadDirectoryChangesWatcher> {
         let (cmd_tx, cmd_rx) = unbounded();
 
-        let wakeup_sem =
-            unsafe { synchapi::CreateSemaphoreW(ptr::null_mut(), 0, 1, ptr::null_mut()) };
-        if wakeup_sem.is_null() || wakeup_sem == INVALID_HANDLE_VALUE {
+        let wakeup_sem = unsafe { CreateSemaphoreW(ptr::null_mut(), 0, 1, ptr::null_mut()) };
+        if wakeup_sem == 0 || wakeup_sem == INVALID_HANDLE_VALUE {
```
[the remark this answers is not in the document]
GamePad64: Yes, there are none. And HANDLE is just `isize` alias.

notify/src/windows.rs
kangalio: I sometimes use a custom prelude module for that 

```rust
mod win  {
    pub use windows_sys::Win32::Foundation::*;
    pub use windows_sys::Win32::Storage::*;
    pub use windows_sys::Win32::Threading::*;
    pub use windows_sys::Win32::WindowsProgramming::*;
    pub use windows_sys::Win32::IO::*;
}

// ...
handle = win::CreateFileW(
    encoded_path.as_ptr(),
    win::FILE_LIST_DIRECTORY,
    win::FILE_SHARE_READ | win::FILE_SHARE_DELETE | win::FILE_SHARE_WRITE,
    ptr::null_mut(),
    win::OPEN_EXISTING,
    win::FILE_FLAG_BACKUP_SEMANTICS | win::FILE_FLAG_OVERLAPPED,
    0,
);
```

Issue #412, opened 2022-06-04
Large scale watching of paths appears to drop/miss events


chipsenkbeil:
## System details

- Rust version (if building from source): `rustc --version`: rustc 1.60.0 (7737e0b5c 2022-04-04)
- Notify version (or commit hash if building from git): 5.0.0-pre.15

### Mac

M1 Macbook Air from 2020.

<img width="228" alt="image" src="https://user-images.githubusercontent.com/2481802/172024821-582338f8-fbc0-4557-a7a6-d079765e2110.png">

### Windows

Traditional Windows 11 machine.

<img width="490" alt="image" src="https://user-images.githubusercontent.com/2481802/172024790-68c12700-7ac0-4ef9-a865-afc471ea6e70.png">

### Linux

Running Alpine Linux using Mac Parallels on an M1 Mac, which means Alpine is using ARM.

<img width="795" alt="image" src="https://user-images.githubusercontent.com/2481802/172024963-a102b6cc-715c-4daa-a453-0dc96a33dfaf.png">

## What you did (as detailed as you can)

Example repo where you can run test: https://github.com/chipsenkbeil/notify-stress-test

1. Create a watcher using `notify::recommended_watcher`
2. Have events get sent from the watcher handler out of the thread using `std::sync::mpsc::Sender`
3. Continually receive events on a separate thread using `std::sync::mpsc::Receiver`
4. Create a large number of paths to watch (files or directories) and watch each path using the same watcher
5. Perform something within each path (modify a file, add a file to a directory, etc)
6. Tick off each path with an event received from the `std::sync::mpsc::Receiver`

## What you expected

Path modifications or other events show up for watched paths at scale. For example, if watching 1500 individual files, modifying each file would result in some event being captured and passed along.

## What happened

With enough different paths being watched, events start to be missing or dropped. For example, watching 1500 individual files, modifying each file, 246 paths were never reported as modified.

```
❯ cargo test
   Compiling notify-stress-test v0.1.0 (/Users/senkwich/projects/notify-stress-test)
    Finished test [unoptimized + debuginfo] target(s) in 0.80s
     Running unittests (target/debug/deps/notify_stress_test-af3191235b8b1ce0)

running 1 test
test tests::stress_test ... FAILED

failures:

---- tests::stress_test stdout ----
thread 'tests::stress_test' panicked at 'assertion failed: `(left == right)`
  left: `246`,
 right: `0`: 246/1500 file paths not modified', src/lib.rs:88:9
[...]

PR #467, opened 2023-01-28
implement passive debouncer without blocking

### LINKED ISSUE #466: `notify-debouncer-mini` always does busy waiting
Previously the notify v4 were doing passive waiting with at least inotify(and it was falling back to polling each X ms), meaning that it was woken up only when something was actually changing, with notify v5 we have constant CPU draining when using the `notify-debouncer-mini`.

How could we avoid such a thing and do passive waiting instead like it was by default before? From reading the documentation it's not clear to me how to avoid that, maybe we should use the notify directly here?

0xpr03:
Remove active polling and introduce a passive waiting method that can also shutdown everything.

This fixes #466

### DIFF
--- a/notify-debouncer-mini/src/lib.rs
+++ b/notify-debouncer-mini/src/lib.rs
@@ -1,20 +1,21 @@
-//! Debouncer for notify
+//! Debouncer for [notify](https://crates.io/crates/notify). Filters incoming events and emits only one event per timeframe per file.
 //!
 //! # Installation
 //!
 //! ```toml
 //! [dependencies]
-//! notify-debouncer-mini = "0.2.0"
+//! notify-debouncer-mini = "0.4.0"
 //! ```
 //! In case you want to select specific features of notify,
 //! specify notify as dependency explicitely in your dependencies.
 //! Otherwise you can just use the re-export of notify from debouncer-mini.
 //! ```toml
-//! notify-debouncer-mini = "0.2.0"
+//! notify-debouncer-mini = "0.4.0"
 //! notify = { version = "..", features = [".."] }
 //! ```
 //!  
 //! # Examples
+//! See also the full configuration example [here](https://github.com/notify-rs/notify/blob/main/examples/debouncer_mini_custom.rs).
 //!
 //! ```rust,no_run
 //! # use std::path::Path;
@@ -22,21 +23,21 @@
 //! use notify_debouncer_mini::{notify::*,new_debouncer,DebounceEventResult};
 //!
 //! # fn main() {
-//!     // setup initial watcher backend config
-//!     let config = Config::default();
+//!   // Select recommended watcher for debouncer.
+//!   // Using a callback here, could also be a channel.
+//!   let mut debouncer = new_debouncer(Duration::from_secs(2), |res: DebounceEventResult| {
+//!       match res {
+//!           Ok(events) => events.iter().for_each(|e|println!("Event {:?} for {:?}",e.kind,e.path)),
+//!           Err(e) => println!("Error {:?}",e),
+//!       }
+//!   }).unwrap();
 //!
-//!     // Select recommended watcher for debouncer.
-//!     // Using a callback here, could also be a channel.
-//!     let mut debouncer = new_debouncer(Duration::from_secs(2), None, |res: DebounceEventResult| {
-//!         match res {
-//!             Ok(events) => events.iter().for_each(|e|println!("Event {:?} for {:?}",e.kind,e.path)),
-//!             Err(errors) => errors.iter().for_each(|e|println!("Error {:?}",e)),
-//!         }
-//!     }).unwrap();
-//!
-//!     // Add a path to be watched. All files and directories at that path and
-//!     // below will be monitored for changes.
-//!     debouncer.watcher().watch(Path::new("."), RecursiveMode::Recursive).unwrap();
+//!   // Add a path to be watched. All files and directories at that path and
+//!   // below will be monitored for changes.
+//!   debouncer.watcher().watch(Path::new("."), RecursiveMode::Recursive).unwrap();
+//! 
+//!   // note that dropping the debouncer (as will happen here) also ends the debouncer
+//!   // thus this demo would need an endless loop to keep running
 //! # }
 //! ```
 //!
@@ -52,15 +53,12 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     path::PathBuf,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc, Mutex,
-    },
+    sync::mpsc::{RecvTimeoutError, Sender},
     time::{Duration, Instant},
 };
 
 pub use notify;
-use notify::{Error, ErrorKind, Event, RecommendedWatcher, Watcher};
+use notify::{Error, Event, RecommendedWatcher, Watcher};
 
 /// The set of requirements for watcher debounce event handling functions.
 ///
@@ -81,8 +79,8 @@ use notify::{Error, ErrorKind, Event, RecommendedWatcher, Watcher};
 ///                     println!("Event {:?} for path {:?}",event.kind,event.path);
 ///                 }
 ///             },
-///             // errors are batched, so you get either events or errors, probably both per debounce tick (two calls)
-///             Err(errors) => errors.iter().for_each(|e|println!("Got error {:?}",e)),
+///             // errors are immediately reported
+///             Err(error) => println!("Got error {:?}",error),
 ///         }
 ///     }
 /// }
@@ -90,6 +88,55 @@ use notify::{Error, ErrorKind, Event, RecommendedWatcher, Watcher};
 pub trait DebounceEventHandler: Send + 'static {
     /// Handles an event.
     fn handle_event(&mut self, event: DebounceEventResult);
+}
+
+/// Config for debouncer-mini
+/// ```rust
+/// # use std::time::Duration;
+/// use notify_debouncer_mini::Config;
+/// let backend_config = notify::Config::default();
+///
+/// let config = Config::default().with_timeout(Duration::from_secs(1)).with_batch_mode(true)
+///     .with_notify_config(backend_config);
+/// ```
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct Config {
+    timeout: Duration,
+    batch_mode: bool,
+    notify_config: notify::Config,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            timeout: Duration::from_millis(500),
+            batch_mode: true,
+            notify_config: notify::Config::default(),
+        }
+    }
+}
+
+impl Config {
+    /// Set timeout
+    ///
+    /// Timeout is the amount of time after which a debounced event is emitted or a continuous event is send, if there still are events incoming for the specific path.
+    pub fn with_timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = timeout;
+        self
+    }
+    /// Set batch mode
+    ///
+    /// When `batch_mode` is enabled, events may be delayed (at most 2x the specified timout) and delivered with others.
+    /// If disabled, all events are delivered immediately when their debounce timeout is reached.
+    pub fn with_batch_mode(mut self, batch_mode: bool) -> Self {
+        self.batch_mode = batch_mode;
+        self
+    }
+    /// Set [`notify::Config`] for the backend
+    pub fn with_notify_config(mut self, notify_config: notify::Config) -> Self {
+        self.notify_config = notify_config;
+        self
+    }
 }
 
 impl<F> DebounceEventHandler for F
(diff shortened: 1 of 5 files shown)


0xpr03:
cc @kchibisov

notify-debouncer-mini/src/lib.rs
```diff
@@ -253,87 +309,81 @@ impl<T: Watcher> Debouncer<T> {
...
+                        }
+                    },
+                    None => match rx.recv() {
+                        Ok(InnerEvent::NotifyEvent(e)) => data.push(e),
+                        Ok(InnerEvent::Shutdown) => run = false,
+                        Err(_) => run = false,
+                    }
+                }
+                
+                // // drain events channel
```
0xpr03, line 351: I'm not sure if we want this, it certainly allows us to receive rapid fired events.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
+    #[inline]
+    pub fn next_tick(&self) -> Option<Duration> {
+        if self.has_tick_data() {
+            let next_run = self.oldest_update.checked_add(self.timeout)
+                .map_or(self.timeout, |v|v.saturating_duration_since(Instant::now()));
+            Some(next_run)
+        } else {
+            None
+        }
+    }
```
kchibisov, line 204: The code looks a bit more complicate than it should. I'd suggest to:

Compute the next debounce time in `Instant` once you receive an event, so whether you should debounce or not. The computation is simply `Instant::now() + self.timeout`(that can't fail). You're already half doing so, but you store `oldest_update` but should store `debounce_deadline`.

Then the total code will look like:

```rust
pub fn next_tick(&self) -> Option<Duration> {
  self.debounce_deadline.map(|deadline| deadline.saturating_duration_since(Instant::now())
}
```

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
+
     /// Retrieve a vec of debounced events, removing them if not continuous
+    ///
+    /// Updates the internal tracker for the next tick
     pub fn debounced_events(&mut self) -> Vec<DebouncedEvent> {
         let mut events_expired = Vec::with_capacity(self.d.len());
         let mut data_back = HashMap::with_capacity(self.d.len());
         // TODO: perfect fit for drain_filter https://github.com/rust-lang/rust/issues/59618
+        // reset oldest update
+        self.oldest_update = Instant::now();
```
kchibisov, line 220: While this doesn't have any implication, with the logic when you don't `Option` for future deadline that looks error prone. Would suggest to simply set to `None` and rewrite the computation around storing the future deadline.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
-
-    /// Add an error entry to re-send later on
-    pub fn add_error(&mut self, e: crate::Error) {
-        self.e.push(e);
-    }
-
     /// Add new event to debouncer cache
-    pub fn add_event(&mut self, e: Event) {
+    #[inline(always)]
+    fn add_event(&mut self, e: Event) {
```
kchibisov, line 241: You can clearly write `event` instead of just `e`.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
+        if self.d.is_empty() {
+            self.oldest_update = time;
+        }
         for path in e.paths.into_iter() {
             if let Some(v) = self.d.get_mut(&path) {
-                v.update = Instant::now();
+                v.update = time;
             } else {
-                self.d.insert(path, EventData::new_any());
+                self.d.insert(path, EventData::new_any(time));
```
kchibisov, line 250: Pretty same comment about the one character variables. We're not writing `C` in the end and have plenty of space to write proper descriptive names.

notify-debouncer-mini/src/lib.rs
```diff
@@ -342,12 +382,9 @@ pub fn new_debouncer_opt<F: DebounceEventHandler, T: Watcher>(
 /// Short function to create a new debounced watcher with the recommended debouncer.
 ///
 /// Timeout is the amount of time after which a debounced event is emitted or a continuous event is send, if there still are events incoming for the specific path.
-///
-/// If tick_rate is None, notify will select a tick rate that is less than the provided timeout.
 pub fn new_debouncer<F: DebounceEventHandler>(
     timeout: Duration,
```
kchibisov, line 386: The docs for this method could say that you invoke the `new_debouncer_opt` with the default `Config`, thus you can point to it.

notify-debouncer-mini/src/lib.rs
kchibisov: I don't think that it'll provide any value? The "rapid fire" events will be handled by `recv_timeout` anyway just fine.

notify-debouncer-mini/src/lib.rs
```diff
@@ -253,87 +289,91 @@ impl<T: Watcher> Debouncer<T> {
...
+                        Ok(InnerEvent::NotifyEvent(e)) => match e {
+                            Ok(v) => data.add_event(v),
+                            Err(e) => event_handler.handle_event(Err(e)),
+                        },
+                        Ok(InnerEvent::Shutdown) => run = false,
+                        Err(_) => run = false,
+                    },
+                }
+                // only debounce if we actually passed a threshold
+                if !debounce {
```
kchibisov, line 338: you can dispatch from the `RecvTimeoutError::Timeout` directly, since that's exactly what you're doing anyway.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
+enum InnerEvent {
+    NotifyEvent(Result<Event, Error>),
+    Shutdown,
+}
 
-#[derive(Default)]
 struct DebounceDataInner {
+    /// Path -> Event data
     d: HashMap<PathBuf, EventData>,
+    /// timeout used to compare all events against, config
```
kchibisov, line 175: I always questioned what is the motivation to remove any information about the events? For example we in alacritty don't want to handle `Removed` events at all to not try to load from the removed path. But that way when you map everything to any event the information for us is sort of discarded. Might forward the events untouched all-together? Will simplify the logic and you can just simply `move` the vector into users.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
     timeout: Duration,
-    e: Vec<crate::Error>,
+    /// last debounce run
+    last_run: Instant,
```
kchibisov, line 178: This value is never read.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
 
 impl<T: Watcher> Debouncer<T> {
     /// Stop the debouncer, waits for the event thread to finish.
     /// May block for the duration of one tick_rate.
     pub fn stop(mut self) {
-        self.set_stop();
+        self.send_stop();
         if let Some(t) = self.debouncer_thread.take() {
             let _ = t.join();
         }
```
kchibisov, line 270: Same one char variables.

notify-debouncer-mini/src/lib.rs
```diff
@@ -253,87 +289,91 @@ impl<T: Watcher> Debouncer<T> {
 impl<T: Watcher> Drop for Debouncer<T> {
     fn drop(&mut self) {
         // don't imitate c++ async futures and block on drop
-        self.set_stop();
+        self.send_stop();
```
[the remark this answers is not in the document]
0xpr03: Dropping won't block and thus wait for the thread to finish. Stop does.

notify-debouncer-mini/src/lib.rs
```diff
@@ -253,87 +289,91 @@ impl<T: Watcher> Debouncer<T> {
...
-                let mut lock = data_c.lock().expect("Can't lock debouncer data!");
-                send_data = lock.debounced_events();
-                errors = lock.errors();
-            }
-            if send_data.len() > 0 {
-                event_handler.handle_event(Ok(send_data));
-            }
-            if errors.len() > 0 {
-                event_handler.handle_event(Err(errors));
+        .spawn(move || {
```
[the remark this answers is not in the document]
0xpr03: I see no reason for that. The thread gets a very obvious name by which you know where it comes from. Everything else just bloats the config for nothing.

notify-debouncer-mini/src/lib.rs
```diff
@@ -160,88 +158,126 @@ pub struct DebouncedEvent {
...
     pub fn stop_nonblocking(self) {
-        self.set_stop();
+        self.send_stop();
     }
 
-    fn set_stop(&self) {
-        self.stop.store(true, Ordering::Relaxed);
+    fn send_stop(&self) {
+        // send error just means that it is stopped, can't do much else
+        let _ = self.stop_channel.send(InnerEvent::Shutdown);
```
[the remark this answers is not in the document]
0xpr03: One consumes self, the other doesn't, required for thread joining.

notify-debouncer-mini/src/lib.rs
0xpr03: If you want specific events, use notify. If you only need information that something changed, use the mini-debouncer. Which is what most people do: Recompute things (assets, binaries,..) when the files change. Debouncing with all event details won't make much sense, you'll just receive a Vec of changes for each path and then have to look through that in order, while the data you got from debouncing might already be wrong the moment you receive them, so concluding anything based on the Vec is wrong. Flattening events to one (create + delete) is something we did, but removed for various problems, as documented in the upgrading guide. Moreover you can't rely on events to not miss out on events, thus you shouldn't ever rely on events for the actual state of files #412

notify-debouncer-mini/src/lib.rs
0xpr03: Done, if we don't handle the possibility of rapid fired events, it makes sense.

notify-debouncer-mini/src/lib.rs
0xpr03: Yeah that's a good idea.

notify-debouncer-mini/src/lib.rs
0xpr03: Changed

notify-debouncer-mini/src/lib.rs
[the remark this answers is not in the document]
kchibisov: But that's the point to clean up correctly, since you want to join on the thread? Otherwise you might exit before you drop the thread.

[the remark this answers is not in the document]
0xpr03: Hm, I'm on the fence about this. The previous event loop would definitely hang for at most `timeout/4`, so this was just to prevent blocking on drop for no good reason.

The new passive debouncer shouldn't really block much, so we could join anyway. But with that the question arises: Do we even want to collect that thread ? We don't really need to, and notify has to spawn various threads for inotify anyway..

notify-debouncer-mini/src/lib.rs
kchibisov: Hm, but at this point, do we really need timestamps from `EventData`? What is the use case for them? Could just return `HashSet<PathBuf>` at this point?

notify-debouncer-mini/src/lib.rs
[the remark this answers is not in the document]
kchibisov: But you have option to drop without any blocking at all with `stop_nonblocking`? I'd assume that watcher will also join and properly reap the inotify? Otherwise you'll leak the inotify nodes, which is not really good, since you might have an on-demand debouncer spawninig, or do some reconfigurations.

Doing proper resource management is what I'd expect from the `drop`(though, if you don't shutdown right away, you'll likely handle that just about fine).

notify-debouncer-mini/src/lib.rs
0xpr03: We don't return EventData, that is just for internal tracking. Otherwise we would report events even if they haven't timed out, so you may be trying to read the config while it's still changing. But if you want to, you can still handle `AnyContinuous`, in case you want to handle the fact something is written for longer than the timeout.

notify-debouncer-mini/src/lib.rs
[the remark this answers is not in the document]
0xpr03: We won't leak anything, we can just decide whether or not to wait for the event loop to finish. We don't event need to create a join handle, we can just start the thread and let it end on its own, it should definitely end once a shutdown command is send, otherwise it'll not end anyway, we our drop will just lock up.

0xpr03:
CC @dfaust because we'll have to retrofit this for the debouncer-full too (especially the semver breaking config change, to future proof additional config values)

PR #470, opened 2023-02-13
Use INotifyWatcher as recommended watcher for Android

tiann:


0xpr03:
Does this require any additional requirements on android over the polling approach ?

0xpr03:
thanks!

PR #471, opened 2023-02-14
Fix a lint warning in examples

y-yagi:
Currently, some examples have the following warning when running.

```
warning: for loop over a `Result`. This is more readably written as an `if let` statement
  --> examples/debounced_full_custom.rs:29:18
   |
29 |         for e in events {
   |                  ^^^^^^
   |
   = note: `#[warn(for_loops_over_fallibles)]` on by default
help: to check pattern in a loop use `while let`
   |
29 |         while let Ok(e) = events {
   |         ~~~~~~~~~~~~~ ~~~
help: consider using `if let` to clear intent
   |
29 |         if let Ok(e) = events {
   |         ~~~~~~~~~~ ~~~
```

This fixed those files as per the linter's suggestion.

### DIFF
--- a/examples/debounced.rs
+++ b/examples/debounced.rs
@@ -27,9 +27,11 @@ fn main() {
         .unwrap();
 
     // print all events, non returning
-    for events in rx {
-        for e in events {
-            println!("{:?}", e);
+    for inner in rx {
+        if let Ok(events) = inner {
+            for event in events {
+                println!("{:?}", event);
+            }
         }
     }
 }
--- a/examples/debounced_full_custom.rs
+++ b/examples/debounced_full_custom.rs
@@ -25,9 +25,11 @@ fn main() {
         .watch(Path::new("."), RecursiveMode::Recursive)
         .unwrap();
     // print all events, non returning
-    for events in rx {
-        for e in events {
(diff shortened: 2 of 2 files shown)


examples/debounced.rs
```diff
@@ -28,7 +28,7 @@ fn main() {
 
     // print all events, non returning
     for events in rx {
-        for e in events {
+        if let Ok(e) = events {
             println!("{:?}", e);
         }
     }
```
JohnTitor, line 34: I guess the original intention is:
```rs
    for inner in rx {
        if let Ok(events) = inner {
            for event in events {
                println!("{:?}", event);
            }
        }
    }
```

examples/debounced_full_custom.rs
```diff
@@ -26,7 +26,7 @@ fn main() {
         .unwrap();
     // print all events, non returning
     for events in rx {
-        for e in events {
+        if let Ok(e) = events {
```
JohnTitor, line 29: Ditto

examples/debounced.rs
y-yagi: Oh, I misunderstood the original intention. I fixed. Thanks!

PR #473, opened 2023-03-21
Replace `windows-sys` with manual bindings

Jake-Shadle:
This replaces the dependency on `windows-sys` with manual bindings to the win32 functionality used by this crate.

The rationale for this change is that these bindings will **never meaningfully change** so it doesn't make sense to use an external dependency, especially in light of https://github.com/microsoft/windows-rs/issues/1720, which is why the currently published version of this crate actually ends up pulling in 2 version of windows-sys since the published version depends on 0.42, but the mio dependency uses 0.45. And as shown in #472, the minor version changes in windows-sys are noise, and only meaningful in the negative impact.

### DIFF
--- a/notify/src/windows.rs
+++ b/notify/src/windows.rs
@@ -20,6 +20,11 @@ use std::slice;
 use std::sync::{Arc, Mutex};
 use std::thread;
 
+/// We do our own minimal bindings to the win32 API here with only the things we
+/// need. This means we can avoid a dependency on the `windows-sys` crate and its
+/// version churn, and the `winapi` due to its maintenance status.
+///
+/// These APIs will never change, though they may be supplanted in the future.
 #[allow(non_camel_case_types, non_snake_case)]
 mod bindings {
     pub const INFINITE: u32 = 4294967295;
--- a/notify/src/windows.rs
+++ b/notify/src/windows.rs
@@ -19,23 +19,133 @@ use std::ptr;
 use std::slice;
 use std::sync::{Arc, Mutex};
 use std::thread;
-use windows_sys::Win32::Foundation::{
-    CloseHandle, ERROR_OPERATION_ABORTED, HANDLE, INVALID_HANDLE_VALUE, WAIT_OBJECT_0,
-};
-use windows_sys::Win32::Storage::FileSystem::{
-    CreateFileW, ReadDirectoryChangesW, FILE_ACTION_ADDED, FILE_ACTION_MODIFIED,
-    FILE_ACTION_REMOVED, FILE_ACTION_RENAMED_NEW_NAME, FILE_ACTION_RENAMED_OLD_NAME,
-    FILE_FLAG_BACKUP_SEMANTICS, FILE_FLAG_OVERLAPPED, FILE_LIST_DIRECTORY,
-    FILE_NOTIFY_CHANGE_ATTRIBUTES, FILE_NOTIFY_CHANGE_CREATION, FILE_NOTIFY_CHANGE_DIR_NAME,
-    FILE_NOTIFY_CHANGE_FILE_NAME, FILE_NOTIFY_CHANGE_LAST_WRITE, FILE_NOTIFY_CHANGE_SECURITY,
-    FILE_NOTIFY_CHANGE_SIZE, FILE_NOTIFY_INFORMATION, FILE_SHARE_DELETE, FILE_SHARE_READ,
-    FILE_SHARE_WRITE, OPEN_EXISTING,
-};
-use windows_sys::Win32::System::Threading::{
-    CreateSemaphoreW, ReleaseSemaphore, WaitForSingleObjectEx,
-};
-use windows_sys::Win32::System::WindowsProgramming::INFINITE;
-use windows_sys::Win32::System::IO::{CancelIo, OVERLAPPED};
+
+#[allow(non_camel_case_types, non_snake_case)]
+mod bindings {
+    pub const INFINITE: u32 = 4294967295;
+    pub const INVALID_HANDLE_VALUE: isize = -1;
+    pub type HANDLE = isize;
+
+    pub type FILE_ACCESS_FLAGS = u32;
+    pub const FILE_LIST_DIRECTORY: FILE_ACCESS_FLAGS = 1;
+
+    pub type FILE_SHARE_MODE = u32;
+    pub const FILE_SHARE_DELETE: FILE_SHARE_MODE = 4;
+    pub const FILE_SHARE_READ: FILE_SHARE_MODE = 1;
+    pub const FILE_SHARE_WRITE: FILE_SHARE_MODE = 2;
+
+    pub type BOOL = i32;
+
+    #[repr(C)]
+    pub struct SECURITY_ATTRIBUTES {
+        pub nLength: u32,
+        pub lpSecurityDescriptor: *mut std::ffi::c_void,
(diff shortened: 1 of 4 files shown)


0xpr03:
Note: we just accepted #457 to switch towards windows-sys from winapi (and #472 to update windows-sys)

notify/src/windows.rs
```diff
@@ -19,23 +19,133 @@ use std::ptr;
...
-};
-use windows_sys::Win32::System::Threading::{
-    CreateSemaphoreW, ReleaseSemaphore, WaitForSingleObjectEx,
-};
-use windows_sys::Win32::System::WindowsProgramming::INFINITE;
-use windows_sys::Win32::System::IO::{CancelIo, OVERLAPPED};
+
+#[allow(non_camel_case_types, non_snake_case)]
+mod bindings {
+    pub const INFINITE: u32 = 4294967295;
```
0xpr03, line 25: my first reaction to this line was: this is why I wouldn't want doing the bindings myself, magic numbers from an OS for which I have no way of validating them, and if they're getting stale, it will take years till people notify (heh) that we're the ones getting it wrong

notify/src/windows.rs
```diff
@@ -19,23 +19,133 @@ use std::ptr;
...
-    FILE_SHARE_WRITE, OPEN_EXISTING,
-};
-use windows_sys::Win32::System::Threading::{
-    CreateSemaphoreW, ReleaseSemaphore, WaitForSingleObjectEx,
-};
-use windows_sys::Win32::System::WindowsProgramming::INFINITE;
-use windows_sys::Win32::System::IO::{CancelIo, OVERLAPPED};
+
+#[allow(non_camel_case_types, non_snake_case)]
+mod bindings {
```
0xpr03, line 24: Please add a comment linking to this PR / adding some context that not using a 3rd party crate is a known tradeoff. (Preventing future PRs or issues..)

0xpr03:
I'm currently tending towards reverting #457 and simply going back to winapi. It may not be developed actively, but a) the (x86_64) winapi probably won't change until the heat death of the universe, b) we're not using anything special and c) that means winapi users converge towards one version.

0xpr03:
ok let's switch this around: if you want to have this, I'm willing to add a feature flag for this, let's call it `slim_windows`, and if that's enabled, the windows-sys crate isn't used

Jake-Shadle:
That would work! Traveling tomorrow so might not have it for a bit.

Jake-Shadle:
So unfortunately due to how features work I don't think that exact scenario works, but I've made `windows-sys` an optional, default crate, and added a `slim_windows` feature, if either that feature is enabled or `windows-sys` is disabled then the manual bindings are used instead of the windows-sys ones.

notify/src/windows.rs
```diff
@@ -19,23 +19,34 @@ use std::ptr;
...
+    pub use windows_sys::Win32::System::Threading::{
+        CreateSemaphoreW, ReleaseSemaphore, WaitForSingleObjectEx,
+    };
+    pub use windows_sys::Win32::System::WindowsProgramming::INFINITE;
+    pub use windows_sys::Win32::System::IO::{CancelIo, OVERLAPPED};
+}
+#[cfg(all(feature = "windows-sys", not(feature = "slim_windows")))]
+use sys::*;
+
+#[cfg(any(not(feature = "windows-sys"), feature = "slim_windows"))]
```
0xpr03, line 46: This is sadly a slippery slope. People use no-default-features to specify kqueue as macos backend. By this definition they automatically get the slim backend, which is unintentional (and got me wondering why the CI passed in the first place).

0xpr03: I've already had a working change to make it a requirement, but then we'd have to bump the major version as everyone has to migrate their features. So that's not really a solution.

Jake-Shadle: I originally had it on just whether they had the slim feature enabled, but if they disable default features and don't enable slim it would fail to compile, but if you're OK with that I can do that instead.

GamePad64:
`windows-sys` is a pretty slim wrapper itself, and its versioning is organized the way it is to provide an ability to use different versions of it in different nodes of dependency tree without any negative circumstances.

We can just pin `windows-sys` to a specific version and forget about it. `mio` can use another version of it and it would be pretty fine.

Jake-Shadle:
`windows-sys` is by far the largest dependency most crates will have in terms of download and disk space, and even with only the needed features enabled, this crate uses a small fraction of the enabled APIs, making the dependency wasteful. Pinning would be even worse as then you are guaranteeing that multiple versions will be used in even a moderately sized dependency graph.

GamePad64:
There is nothing wrong of having a crate installed multiple times. But there is a problem of having to write the bindings by yourself.
Just imagine, that you want to use some new function from WinAPI, that you have never used before in your project. So, you have to create a binding by yourself, and then test it by hand. So you have to think about the correctness of boilerplate code instead of business logic issues.
Also, the willingness of new contributions by the community would be much lower, bc no one likes writing boilerplate code by themselves :)

`windows-sys`:
Positive:
- Reduces boilerplate code
- Reduces TTM value
- Ensures correctness, by being autogenerated from win32metadata, which is used in most win32 bindings by Microsoft.
- Actively maintained
Negative:
- About 2MB of gzipped code (srsly?)
- Can have breaking changes between versions (0.* version numeration)

I strongly advise against pulling the bindings in the project code and taking the burden of maintaining the bindings by the project contributors.

0xpr03:
Ok, so this would still be a [breaking change](https://github.com/search?q=notify+lang%3Atoml&type=code), thus requiring a v6 release. We would have to select windows-sys when no feature is specified?

Jake-Shadle:
Yes, it's a breaking change for those who use `default-features = false`, but the mitigation for that is the explicit `compile_error!` to tell the user what went wrong.

0xpr03:
Hm, that's a fair trade I guess ?

Jake-Shadle:
IMO yes, though there could be an issue if a crate depends on notify with `default-features = false`, then doesn't compile for Windows, does a publish, and then gets transitively depended on by a crate/project that does compile for Windows.

GamePad64:
There should be a simple process of updating the bindings, prior to merging this PR.

0xpr03:
> simple process of updating the bindings

What do you mean with that ? As it currently stands, this will be v6

0xpr03:
On one hand it's not really a problem to do a v6 for this. But on the other side, doing a v6 only for a feature that, as of now, exactly one Person uses is kind of weird.

riverar:
> The rationale for this change is that these bindings will never meaningfully change

This assertion continues to be untrue, like I've mentioned before in other repositories.

Is there a demonstrable problem here? Doesn't appear version churn was an issue for this crate.

GamePad64:
> guaranteeing multiple versions since other crates have updated to new versions.

So what, exactly? This isn't a problem in Rust ecosystem, as using multiple versions of the same library, when a major version is updated, is a Rust feature, not a bug.
The issue you are trying to solve is miniscule, compared to project contributors' burden to maintain the bindings and regenerate them if they need a new WinAPI function.

The approach you are proposing (`slim-windows` feature) just makes it more difficult to maintain `notify-rs` and gives nothing in return.

What real benefit do you want to bring by this change?

Jake-Shadle:
Closing this since this thread is unproductive. Sorry for the headache @0xpr03 .

PR #479, opened 2023-04-26
Update `windows-sys` to `0.48`

a1phyr:


0xpr03:
Thanks, but I think we'll wait for now. See #473. I'm waiting for this to develop further than some "cosmetic" changes of where the imports are.

complexspaces:
Hi there, has there been any more thinking on this point in the last few months? It seems like the `windows` ecosystem totals to providing 3 options now:
- `windows-sys`, like what `notify` is currently using.
- `windows-bindgen`, which is what the previously attempted PR wanted to introduce in some form.
- `windows-target` for manually declaring bindings but still getting proper import resolution.

We're currently trying to consider a `windows` ecosystem upgrade and Notify is in-tree for us.

0xpr03:
In general we can upgrade to 0.48  
What do I do want to prevent is having weekly upgrades for changes in the windows crates that are purely syntactical (the bindings themselves don't change for us).

In terms of code side this shouldn't affect anyone, as LTO should remove everything that isn't used, and the bindings are not big.

There is probably also a debate to be had about windows-sys vs windows-bindgen vs windows of notify vs debouncer-full

Issue #458, opened 2022-12-12
ModifyKind does not derive Copy


jasonheath:
Apologies if I should ask this in a different place but I think the issues section is correct for this project.  I've noticed that ModifyKind doesn't derive Copy. Based on a quick visual inspection I believe that all of its peer enums and the enums it makes use of derive Copy.  Is there a reason it doesn't?  If there isn't I can submit a pull request that derives copy.

0xpr03:
Yeah there is currently no real blocker for deriving copy on top of clone. But it also will make it impossible to remove, if for any reason we should want a bigger sub type / value later on. But I could be too restrictive in favor of unknown unknowns.

Is there a good use case for making it copy ?

dfaust:
> Is there a good use case for making it copy ?

When matching on `ModifyKind`, it is moved into the match and so the event cannot be used anymore.
I believe it should be `Copy` just like the other enums, so that this use case is easier and the API is more uniform.

![grafik](https://user-images.githubusercontent.com/893710/234108431-87bb6991-893b-4ee6-9d0a-81129cf66005.png)

0xpr03:
I would suggest matching over &event. Still, I would merge a PR for this.

Issue #475, opened 2023-04-14
I use notify to watch a path which is mount by nfs，notify didn't work and no error return


co0ontty:
#### What you did (as detailed as you can)

I use notify to watch a path which is mount by nfs，notify didn't work and no error return

#### What you expected

I want recive a error when notify can't watch this path

0xpr03:
You need to use the [PollWatcher](https://docs.rs/notify/latest/notify/poll/struct.PollWatcher.html) for network mounts. See also [stackoverflow](https://stackoverflow.com/a/4231277/3332686).

Notify can't tell whether this will work, as the kernel just doesn't send any events for the NFS mount.

co0ontty:
If possible, can you return an error when the nfs type appears in the directory?

0xpr03:
We don't know when something is NFS. As far as I know, we would have to scan your mounts and detect an NFS system, even across soft- and hard-links. That seems out of scope to me.

PR #480, opened 2023-05-09
Add refined debouncer

dfaust:
This is another take at event debouncing.
Unlike the `mini` debouncer, the `refined` debouncer is much more opinionated.
It tries its best to simplify event handling and in particular can use file system ids to stitch together rename events in case the notification back-end doesn't support cookies.

Features:
* Only emits a single `Rename` event if the rename `From` and `To` events can be matched
* Merges multiple `Rename` events
* Optionally keeps track of the file system IDs all files and stitches rename events together (FSevents, Windows)
* Emits only one `Remove` event when deleting a directory (inotify)
* Doesn't emit duplicate create events
* Doesn't emit `Modify` events after a `Create` event

This PR also removes the `rename` thread from the inotify back-end and simplifies the rename event handling.
Because of that I bumped the version number to 6.0.0.

I haven't done much testing, yet. But it works well on Linux so far.

This PR also implements `Copy` for `EventKind` and `ModifyKind` (#458).
~~And I changed the dependency declarations to use workspace dependencies.~~ (Requires Cargo 1.64)


0xpr03:
I also want to throw another idea at you: I was thinking about whether we could make a "fat" debouncer that also detects filesystems that require usage of pollwatcher. It could then automatically use the pollwatcher backend in addition to the native one. (See #475 ) This would of course mean detecting the added path in the mount section for linux. It would already help with missing events for WSL environments that watch windows host folders

JohnTitor:
Meanwhile, could you split https://github.com/notify-rs/notify/issues/458 as a separate PR? It's essentially unrelated to the debouncer change here and it's worth releasing it as v5.2.0.

dfaust:
> I was thinking about whether we could make a "fat" debouncer ...

@0xpr03 
I like the name "fat" debouncer. "refined" was the best name I could come up with.
I'm not sure how combining the poll watcher with the native watcher would look like,
but there are a few areas where a debouncer and the poll watcher can be integrated closer.
I was thinking about sharing the file ID cache (this is also a reason why I made the `FileIdCache` trait).
Going even further it may be possible to create some sort of "offline" change detection, where events are reported for changes that happened while the application wasn't running.

> could you split #458 as a separate PR?

@JohnTitor 
Sure.

Cargo.toml
```diff
@@ -3,14 +3,11 @@
 members = [
   "notify",
   "notify-debouncer-mini",
-  
+  "notify-debouncer-refined",
```
0xpr03, line 6: let's change this to "notify-debouncer-full"

notify-debouncer-refined/src/lib.rs
```diff
@@ -0,0 +1,788 @@
...
+            Error::new(ErrorKind::Generic(format!(
+                "Failed to calculate tick as {:?}/{}!",
+                timeout, tick_div
+            )))
+        })?,
+    };
+
+    let data_c = data.clone();
+    let stop_c = stop.clone();
+    let thread = std::thread::Builder::new()
```
0xpr03, line 582: if you want to, you can already move ahead of #467 and use a tick-less system - otherwise we could introduce this later on

the idea is to either let a new event wake up the debouncer, or schedule a sleep-timeout if we need another tick (remaining events in the event queue)

I'll merge #467 after this one for the mini debouncer, and can also do the followup for this one, if you want

notify-debouncer-refined/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+name = "notify_debouncer_refined"
+path = "src/lib.rs"
+
+[features]
+default = ["crossbeam"]
+# can't use dep:crossbeam-channel and feature name crossbeam-channel below rust 1.60
```
0xpr03, line 23: I think we can just break the version compatibility here, if we're doing a major release anyway for notify itself

So feel free to just undo the 1.56 compatibility thingy and I may clean up stuff like this later on

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 v4 commits split out to branch `v4_maintenance` starting with `4.0.16`
 
+## notify 6.0.0
+
+- CHANGE: files and directories moved into a watch folder on Linux will now be reported as `rename to` events instead of `create` events
```
0xpr03, line 7: can confirm this

0xpr03 (changes requested):
Looks overall pretty good. A ton more logic in the debouncer now, but this will give a much better user experience.

As noted I would suggest renaming to -debouncer-full.
Fat is also an option, but sounds more negative than it has to.

I would drop 1.56 compatibility and move on, as for the crossbeam things we could really need some functionality from 1.60+

I really like the testing harness.

file-id/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
```
0xpr03, line 1: just to confirm, we don't have to publish this, right ?

notify-debouncer-refined/src/lib.rs
dfaust: I only saw #467 after I finished my PR. I think we should merge it first and then improve it.

notify-debouncer-refined/Cargo.toml
dfaust: I undid the 1.56 compatibility commit and raised the MSRV to 1.60.

file-id/Cargo.toml
dfaust: That's right.
I think there could be some utility in publishing it, but there is also room for improvement.
Currently it only supports the old, low resolution, ids on windows. Newer file systems have file ids with a size 128 bits instead of 64 bits.

examples/Cargo.toml
```diff
@@ -5,9 +5,11 @@ publish = false
 edition = "2021"
 
 [dev-dependencies]
-notify = { version = "5.1.0" }
-notify-debouncer-mini = { version = "0.2.0" }
+notify = { version = "6.0.0", path = "../notify" }
+notify-debouncer-mini = { version = "0.2.0", path = "../notify-debouncer-mini" }
```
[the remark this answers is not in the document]
dfaust: In the long term it would be nice to use workspace dependencies. That would make things simpler. But they require Rust 1.64.

0xpr03:
> I think we should merge it first and then improve it.

If you have something you want to merge pre 6.0 release leave me comment, otherwise I'd publish a v6 tomorrow.

dfaust:
@0xpr03 
Will you update the debouncer-full as well, after merging #467, or should I do that? Please let me know.

I think a new release is great, but as I said, I haven't done much testing, yet.
Maybe we could include a call for testing in the release notes.

0xpr03:
Yeah I'll try to do that as well.

> call for testing 

We could do a RC

dfaust:
> We could do a RC (but this time not for 2 years)

:smile: I'll leave that decision to you.
Personally, since this is the first release of the full debouncer, I would release it as version 0.1.0 and just add a few lines to the release notes saying something like: "This release contains a new debouncer. It should still be considered experimental at this point. Please let us know if you encounter any issues with it."

0xpr03:
Ah yeah I meant RC for v6 of notify. But I guess we'll skip that. And for the debouncer-full a 0.1.0 is fine.

0xpr03:
This got a little bit hairy, apart from me not noticing that the updated versions weren't all correct, crates.io does require publishing file-id. So that's now also up there.

PR #481, opened 2023-05-09
Implement `Copy` for `EventKind` and `ModifyKind`

### LINKED ISSUE #458: ModifyKind does not derive Copy

dfaust:
Closes #458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 v4 commits split out to branch `v4_maintenance` starting with `4.0.16`
 
+## notify 5.2.0
+
+- CHANGE: implement `Copy` for `EventKind` and `ModifyKind` [#458]
+
 ## notify 5.1.0 (2023-01-15)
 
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 v4 commits split out to branch `v4_maintenance` starting with `4.0.16`
 
+## notify 5.2.0
```
0xpr03, line 5: do you want an additional release for this ? could just batch this into the 6.0

0xpr03 (approved):

CHANGELOG.md
0xpr03: agreed

PR #488, opened 2023-05-28
Add `DebouncedEvent` to debouncer-full

dfaust:
Emit events in the debouncer-full as `DebouncedEvent`s, each containing the original notify event and the time at which it occurred.

Having access to the time at which the event occurred is useful when handling situations in which the application writes to the file system and has to ignore the resulting events within a certain time frame.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 v5 maintenance branch is on `v5_maintenance` after `5.2.0`  
 v4 commits split out to branch `v4_maintenance` starting with `4.0.16`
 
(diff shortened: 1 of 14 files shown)


0xpr03:
LGMT, ping me when you're ready

chipsenkbeil:
@dfaust can this get merged? Timestamps would be really handy for my use case (integrating with an editor that writes to watched files). Is there anything blocking this now? 😄

dfaust:
> @dfaust can this get merged?

Yes, I haven't found any issues and haven't made any changes.

Issue #491, opened 2023-06-08
Directory moved from one watched directory to another is not recursively watched


emilycf-cw:
#### System details

- OS: Linux pop-os 6.2.6-76060206-generic
- Rust version: rustc 1.70.0 (90c541806 2023-05-31)
- Notify version: 6.0.0

- Filesystem type and options: ext4
- Linux kernel version: 6.2.6-76060206-generic
- If you're running as a privileged user (root, System): no

#### What you did (as detailed as you can)

<details><summary>Code</summary>
<p>

```rust
use notify::{Config, RecommendedWatcher, RecursiveMode, Watcher};

/// Async, futures channel based event watching
fn main() {
    let mut args = std::env::args();
    let _ = args.next();
    let path_1 = args.next()
        .expect("Argument 1 needs to be a path");
    let path_2 = args.next();

    let (tx, rx) = std::sync::mpsc::channel();

    // Automatically select the best implementation for your platform.
    // You can also access each implementation directly e.g. INotifyWatcher.
    let mut main_watcher = RecommendedWatcher::new(
        move |res| {
            tx.send(res).expect("channel closed");
        },
        Config::default(),
    ).expect("Failed to create watcher");
    // Add a path to be watched. All files and directories at that path and
    // below will be monitored for changes.
    println!("watching {path_1} recursively");
    main_watcher.watch(path_1.as_ref(), RecursiveMode::Recursive).expect("Failed to watch first path");
    if let Some(path_2) = path_2 {
        println!("watching {path_2} recursively");
        main_watcher.watch(path_2.as_ref(), RecursiveMode::Recursive).expect("Failed to watch second path");
    }

    while let Ok(res) = rx.recv() {
        match res {
            Ok(event) => println!("{:?}: {:?}", event.kind, event.paths),
            Err(e) => println!("watch error: {:?}", e),
        }
    }
}
```

</p>
</details> 

In one terminal: `cargo run -- dir-1 dir-2`

In another terminal:
```shell
$ mkdir dir-2/subdir
$ mv dir-2/subdir dir-1
$ touch dir-1/subdir/file
```

#### What you expected

After moving a directory from one watched directory to another recursively watched directory, the moved directory is watched.

Expected output (produced by running the same code, but using `notify` v5.1.0 instead of v6.0.0:
```
watching dir-1 recursively
watching dir-2 recursively
Create(Folder): ["/home/emilycf/rust/notify-playground/dir-2/subdir"]
[...]

PR #498, opened 2023-06-21
fix moved directory not being watched

sebastiantoh:
Fixes https://github.com/notify-rs/notify/issues/491

The bug was introduced in https://github.com/notify-rs/notify/commit/65d6d91ca1e97eeb548e41f8694513b280d49d0a#diff-9663de7aadbfa93d9f87e8c0dc5c94855e59fff314c091caa5478b666e404983R258-R264, which calls `add_watch_by_event` only if the trackers matched. 

Prior to https://github.com/notify-rs/notify/commit/65d6d91ca1e97eeb548e41f8694513b280d49d0a, `add_watch_by_event` was called so long as the `event.mask.contains(EventMask::MOVED_TO)` regardless of whether the trackers matched
https://github.com/notify-rs/notify/blob/29376adabb8477698465a3e454ffe6805824fce4/notify/src/inotify.rs#L252-L303

### DIFF
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -254,14 +254,8 @@ impl EventLoop {
                                         .add_some_path(rename_event.paths.first().cloned())
                                         .add_some_path(path.clone()),
                                     );
-                                } else {
-                                    add_watch_by_event(
-                                        &path,
-                                        &event,
-                                        &self.watches,
-                                        &mut add_watches,
-                                    );
                                 }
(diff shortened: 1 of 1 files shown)


0xpr03:
Hmm just to confirm: This doesn't change the emitted events ?

cc @dfaust

0xpr03:
Besides that: Thanks for investing the time to bisect and fix this!

dfaust:
Thanks @sebastiantoh!
I can confirm the fix.

@0xpr03: No, it only fixes the bug.

PR #499, opened 2023-06-21
Add logging

dfaust:
How do you feel about adding some log statements?

I also added a debounced monitor. This should make debugging a bit easier.


0xpr03:
My only concern would be an additional crate, performance wise it shouldn't matter. It would definitely be kinda breaking for people that didn't expect massive log spam from notify previously.

PR #508, opened 2023-07-15
add readme for examples

0xpr03:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Notify
 
 [![» Crate](https://flat.badgen.net/crates/v/notify)][crate]
-[![» Docs](https://flat.badgen.net/badge/api/docs.rs/df3600)][docs]
+[![» Docs](https://flat.badgen.net/badge/api/docs.rs/df3600)][notify-docs]
 [![» CI](https://flat.badgen.net/github/checks/notify-rs/notify/main)][build]
 [![» Downloads](https://flat.badgen.net/crates/d/notify)][crate]
 [![» Conduct](https://flat.badgen.net/badge/contributor/covenant/5e0d73)][coc]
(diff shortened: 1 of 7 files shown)


0xpr03:
@dfaust could it be that `examples/debouncer_full.rs` and `examples/monitor_debounced.rs` are the same, and we can just delete one of them ?

dfaust:
`monitor_debounced` watches for changes in a directory that the user must provide,
while `debouncer_full` watches a temporary directory and continuously writes to a file and renames it repeatedly.
So they do different things, I think it would be ok to remove `debouncer_full` if you want to cleanup a bit, but keep that in mind.

0xpr03:
Ah yeah it's a little confusing as [cache()](https://github.com/notify-rs/notify/blob/af4c021dc15a3e514a36a4ff5c1d61ee7664d6a6/notify-debouncer-full/src/lib.rs#L540C1-L540C1) is lacking its correct documentation. So it's not exactly clear why I should be able to access this as a normal user.

dfaust:
Yes, the cache feature is poorly documented.
The cache holds the paths and file ids of all watched files and is used to stitch together rename events on platforms that don't support rename cookies.

I wanted to let the user provide their own cache, since they might want to keep a cache with similar data anyway, so they could save a bit of memory. But right now, it is unnecessarily confusing.

One option would be to remove the `watcher()` method and implement `Watcher` for `Debouncer`. Then adding watched paths to the cache could be done automatically.

Maybe we can remove the `cache()` method as well. I was planning on using the custom cache feature in a project of mine, but haven't gotten so far, yet.

0xpr03:
I think it's fine to keep this, but maybe add something that clarifies exactly that:
- not required normally
- useful if you need that cache in your own code

And maybe some additional example for that case. (And also a fixed doc)

0xpr03:
> implement Watcher for Debouncer

I think this would be a good way forward. Especially if we want to add multi-backend support. The main use case would be detecting network filesystems and using the PollWatcher for those. In that case there is not one `watcher()` to return, so it makes no sense to keep that API.

PR #511, opened 2023-07-18
test BSD build

0xpr03:
just a test for #494


0xpr03:
Ok so the test binary creates the problem.

0xpr03:
This still misses a way to run the cross-compilation for netbsd.

Issue #516, opened 2023-08-02
Potential Double Free Issue


kuzeyardabulut:
Hi,
I found a memory-safety/soundness issue in this crate while scanning Rust code for potential vulnerabilities. This PR contains a fix for the issue.

# Issue Description
https://github.com/notify-rs/notify/blob/5f40b83c04b04893a42f91b55e4ceeef3777d47c/notify/src/windows.rs#L281-L310

If a panic!() occurs between the `Box::new()` function and `std::mem::forget`, a double free vulnerability emerges.

# Fix
In Rust, `std::mem::forget` does not actually free the memory, instead it simply allows the memory to leak. This can lead to double free when the data object goes out of scope and its destructor is called automatically. The modification here uses `std::mem::ManuallyDrop` to wrap data. This ensures that data will not be automatically dropped when it goes out of scope, thus avoiding a double free scenario. With `ManuallyDrop`, we explicitly state that the data variable should not be dropped, thus avoiding any potential double free issues.

0xpr03:
Hi, thanks for reporting.
Adding `ManuallyDrop` doesn't seem to be wrong. However I can't find any way this could panic - at least not in a way that it creates a problem. The only thing that could panic is `ReadDirectoryChangesW` itself, which would mean the C-API of windows is panicking, as it's a [direct wrapper](https://docs.rs/windows-sys/0.48.0/windows_sys/Win32/Storage/FileSystem/fn.ReadDirectoryChangesW.html#).

kuzeyardabulut:
Thanks for your answer. For now, there is no possible panic!() here, but it may be important to fix this error for the progress of the project.

I will create a PR. Please let me know if there is any mistake.

PR #517, opened 2023-08-02
Fixing Potential Double Free Issue

kuzeyardabulut:
Hi,
I found a memory-safety/soundness issue in this crate while scanning Rust code for potential vulnerabilities. This PR contains a fix for the issue.

# Issue Description
https://github.com/notify-rs/notify/blob/5f40b83c04b04893a42f91b55e4ceeef3777d47c/notify/src/windows.rs#L281-L310

If a panic!() occurs between the `Box::new()` function and `std::mem::forget`, a double free vulnerability emerges.

Related [Issue](https://github.com/notify-rs/notify/issues/516)


notify/src/windows.rs
```diff
@@ -295,19 +295,16 @@ fn start_read(rd: &ReadData, event_handler: Arc<Mutex<dyn EventHandler>>, handle
...
             Some(handle_event),
         );
 
         if ret == 0 {
             // error reading. retransmute request memory to allow drop.
-            // allow overlapped to drop by omitting forget()
+            // allow overlapped to drop by dropping ManuallyDrop
             let request: Box<ReadDirectoryRequest> = mem::transmute(request_p);
-
+            std::mem::ManuallyDrop::drop(&mut overlapped);
```
0xpr03, line 306: this changes the logic - we dropped it in case ret != 0, now we're dropping it right before a releasesemaphore call on ret == 0

0xpr03 (changes requested):

notify/src/windows.rs
0xpr03: Also please don't name your commit "Update windows.rs"

kuzeyardabulut: > Looks like it's still wrong?

How can i fix it ?

Nemo157: The `drop` call should still be in the `if ret == 0` path (and none in the `else` path), but _after_ the `ReleaseSemaphore` call, so the value is dropped at the same place it would have been previously.

kuzeyardabulut: Thanks!

adamreichold: I would like to add that if `ManuallyDrop` is used, then [`ManuallyDrop::into_inner`](https://doc.rust-lang.org/stable/std/mem/struct.ManuallyDrop.html#method.into_inner) should be used re-arm the `Drop` impl of `Box<OVERLAPPED>`.

Personally, I have to say that using the `mem::forget` does look like the superior solution in this case, from a clarity and therefore safety perspective. Especially since the original issue description

> If a panic!() occurs between the Box::new() function and std::mem::forget, a double free vulnerability emerges.

is wrong: Conceptually, ownership of `overlapped` is passed to `ReadDirectoryChangesW` which is why we need to disarm the `Drop` when that function indicates success. So a panic would have to occur between `ReadDirectoryChangesW` returning and `mem::forget` to open the code to a potential double free, but there is no code in between for `ret != 0`. A panic between `Box::new` and `ReadDirectoryChangesW` would not imply a double-free opportunity.

notify/src/windows.rs
```diff
@@ -295,19 +295,17 @@ fn start_read(rd: &ReadData, event_handler: Arc<Mutex<dyn EventHandler>>, handle
...
             &mut 0u32 as *mut u32, // not used for async reqs
-            &mut *overlapped as *mut OVERLAPPED,
+            (&mut **overlapped) as *mut OVERLAPPED,
             Some(handle_event),
         );
 
         if ret == 0 {
             // error reading. retransmute request memory to allow drop.
-            // allow overlapped to drop by omitting forget()
+            // allow overlapped to drop by omitting drop()
```
danielhenrymantilla, line 304:
```suggestion
            // Because of the error, ownership of `overlapped` was not passed
            // over to `ReadDirectoryChangesW`.
            // So we can claim ownership back
            let _overlapped = ManuallyDrop::into_inner(overlapped);
```

notify/src/windows.rs
```diff
@@ -295,19 +295,17 @@ fn start_read(rd: &ReadData, event_handler: Arc<Mutex<dyn EventHandler>>, handle
...
             let request: Box<ReadDirectoryRequest> = mem::transmute(request_p);
-
             ReleaseSemaphore(request.data.complete_sem, 1, ptr::null_mut());
-        } else {
-            // read ok. forget overlapped to let the completion routine handle memory
-            mem::forget(overlapped);
+            
+            std::mem::ManuallyDrop::drop(&mut overlapped);
```
danielhenrymantilla, line 308:
```suggestion
```

danielhenrymantilla (approved):
The current code is indeed equivalent to the one pre-PR, and robust to some _future_ panic paths that maybe could be introduced between `ReadDirectoryChangesW` succeeding (and thus claiming ownership of the `OVERLAPPED` heap allocation), and the `forget` (to relinquish our ownership of the allocation) happening ✅

I have nonetheless suggested a slight variation which uses `ManuallyDrop::into_inner`, based on what @adamreichold  suggested, since it is a less error-prone API than `drop`, and maps better to what is conceptually happening.

Finally, in this case, since we are "just" dealing with a heap allocation, rather than `ManuallyDrop<Box<…>>`, I'd personally simply use `*mut …`:

```rs
const WIN_FALSE: BOOL = 0;

let overlapped_p = Box::into_raw(Box::new(mem::zeroed::<OVERLAPPED>()));
…
let success = ReadDirectoryChangesW(…);

if success == WIN_FALSE {
    // error reading. retransmute request memory to allow drop.
    // Because of the error, ownership of the `overlapped` alloc was not passed
    // over to `ReadDirectoryChangesW`.
    // So we can claim ownership back.
    let _overlapped_alloc = Box::from_raw(overlapped);
    request = Box::from_raw(request_p);
    ReleaseSemaphore(request.data.complete_sem, 1, ptr::null_mut());
}
```

PR #520, opened 2023-08-08
change license to CC0 for notify and apache/mit for new crates

### LINKED ISSUE #514: (even more) license clarifications: non-standard Artistic-2.0 license
I'm responsible for maintaining packages for the "notify" crate in Fedora Linux, and am also currently in the process of adding "notify-debouncer-mini/full", as more Rust applications start depending on them. Reviewing the project, I noticed that the situation around the license of this project is still a bit unclear.

I asked for advice on the Fedora Project's mailing list for questions like these (included [here](https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/thread/CYDPF2JNJCZWM7LHMF33PWAHLA6Q4AQW/) for reference), and according to a lawyer who specializes in Open Source, there's two potential issues that would require clarifications:

1. The SPDX license in the crate's metadata is not *wrong*, but technically can be considered to be incomplete. Since old code remains "CC0-1.0 only" and new code is "CC0-1.0 OR Artistic-2.0", the SPDX expression for this situation would be "CC0-1.0 AND (CC0-1.0 OR Artistic-2.0)".

2. The license text for the Artistic-2.0 license (originally added in [this commit](https://github.com/notify-rs/notify/commit/a4c0803a7dd48227ce8653fd0e8e38b8e1af2dfa)) has an additional "choice of venue" clause, which is both "unusual" and potentially would disqualify the license terms (combined Artistic-2.0 plus the attached clause) from being considered FOSS in some contexts.

The relevant part from the lawyer's response:

> I think it is potentially problematic from a license policy standpoint
> but the answer isn't immediately obvious. This is a choice of venue
> clause. A small number of licenses generally recognized as FOSS have
> them, but they are controversial in an open source context and I think
> there is a growing recognition that they should probably preclude a
> license from being considered FOSS (I think there was a
> recently-submitted OSI license that had such a feature). Since Fedora
> doesn't have a clear policy on this, you could submit an issue at
> fedora-license-data. We would not treat this as identical to
> `Artistic-2.0` (at least, I don't think we would/should) so if it were
> allowed it would be something like `Artistic-2.0 WITH AdditionRef-NZ-action`
> or something like that, using the recently adopted SPDX 'AdditionRef-'
> construct for custom-defined identifiers for additional terms.
>
> I'd also note that it is unclear whether that term is supposed to
> apply only to the contributions of Félix Saparelli (who apparently is
> in New Zealand) or to all other contributors. In some other context
> that might just be annoying but here it possibly affects whether the
> license is FOSS and it probably affects how you properly represent the
> license as an SPDX expression. I.e. perhaps one would say it is
> `Artistic-2.0 WITH AdditionRef-NZ-action AND Artistic-2.0`. It is
> *not* common at all in open source to see people tacking on choice of
> venue clauses to standard licenses and I think it ought to be viewed
> as inappropriate even if the resultant license should be treated as
> FOSS.

*(from [this mailing list post](https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/message/R5DBCQEBKU66KWIIAAOQ5MH44VEKOY6S/))*

So for now, we will likely opt for using this project under the CC0-1.0 license only (since all code remains available under this license), but it would be great to have a definitive answer, especially concerning the second point.

0xpr03:
Closes #514


0xpr03:
I'm inclined to change the debouncers to the MIT/Apache 2.0 dual license. It is the default for most crates and we still can change this for new stuff, as they're not that old.

cc @passcod @dfaust @decathorpe

decathorpe:
`MIT OR Apache-2.0` would be OK from my point of view as well, but simplifying to CC0-1.0 first is a good idea regardless of the further steps.

dfaust:
I'm fine with any of the licenses, `CC0-1.0`, `MIT`, and `Apache-2.0`.
Personally, would cast my vote for MIT, but then again, I'm not an expert in the field. So please do whatever you think is best.

0xpr03:
If there are no objections, I'll merge this tomorrow.

passcod (approved):

0xpr03:
last commit only moved the notify CC0 license into the notify folder

PR #523, opened 2023-08-12
document more known issues

### LINKED ISSUE #445: docs todo tracking
List of things to fix in the docs

- > platform in **immediate mode**
- > Note that for **recusive** watched
- > If tick_rate is None, **notify** will select a tick rate that is less than the provided timeout.
- > notify-debouncer-mini = "0.2.**0**"
- readme.md for the examples/ folder, looking at the traffic insights
- maybe make it clear that the Config is not consumed by all backends #444 (code comments)
- document WSL2 problems as known problem
- add notice about https://man7.org/linux/man-pages/man7/inotify.7.html
  > The inotify API does not report file accesses and modifications that may occur because of mmap(2), msync(2), and munmap(2).
- document stuff in #412
- add NFS mounts to list of non-working inotify watcher #475

### LINKED ISSUE #503: Watcher isn't Sync with --no-default-features
>> Note: we may want to document and test for this, so it isn't reverted later on with a trait under the impression of usability improvements
>
>I did add a simple test for this in src/lib.rs, where I assert that the `Watcher` trait is object safe. So we can't accidentally remove this feature.

_Originally posted by @erickt in https://github.com/notify-rs/notify/issues/336#issuecomment-858929800_

That only works when not using `std::sync::mpsc::Sender`:

```
$ cargo test -p notify --no-default-features
   Compiling notify v6.0.1 (notify/notify)
error[E0277]: `std::sync::mpsc::Sender<EventLoopMsg>` cannot be shared between threads safely
   --> notify/src/inotify.rs:570:13
    |
570 |     check::<INotifyWatcher>();
    |             ^^^^^^^^^^^^^^ `std::sync::mpsc::Sender<EventLoopMsg>` cannot be shared between threads safely
    |
    = help: within `INotifyWatcher`, the trait `Sync` is not implemented for `std::sync::mpsc::Sender<EventLoopMsg>`
note: required because it appears within the type `INotifyWatcher`
   --> notify/src/inotify.rs:45:12
    |
45  | pub struct INotifyWatcher {
    |            ^^^^^^^^^^^^^^
note: required by a bound in `inotify::inotify_watcher_is_send_and_sync::check`
   --> notify/src/inotify.rs:569:24
    |
569 |     fn check<T: Send + Sync>() {}
    |                        ^^^^ required by this bound in `inotify::inotify_watcher_is_send_and_sync::check`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `notify` due to previous error
```

0xpr03:
closes #445
closes #503

### DIFF
--- a/notify-debouncer-full/src/lib.rs
+++ b/notify-debouncer-full/src/lib.rs
@@ -16,7 +16,7 @@
 //! ```
 //!
 //! In case you want to select specific features of notify,
-//! specify notify as dependency explicitely in your dependencies.
+//! specify notify as dependency explicitly in your dependencies.
 //! Otherwise you can just use the re-export of notify from debouncer-full.
 //!
 //! ```toml
--- a/notify-debouncer-mini/src/lib.rs
+++ b/notify-debouncer-mini/src/lib.rs
@@ -7,7 +7,7 @@
 //! notify-debouncer-mini = "0.4.0"
 //! ```
 //! In case you want to select specific features of notify,
-//! specify notify as dependency explicitely in your dependencies.
+//! specify notify as dependency explicitly in your dependencies.
 //! Otherwise you can just use the re-export of notify from debouncer-mini.
 //! ```toml
 //! notify-debouncer-mini = "0.4.0"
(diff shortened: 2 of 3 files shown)


JohnTitor (approved):
👍

PR #538, opened 2023-10-01
code style improvement 

coswat:
* added `Self` instead of hardcoded impl name -> in notify crate
* organized imports -> in notify crate
* `eprintln!()` macro for Error printing instead of normal println macro -> in file-id crate

### DIFF
--- a/file-id/bin/file_id.rs
+++ b/file-id/bin/file_id.rs
@@ -22,6 +22,6 @@ fn print_file_id(path: &str) {
 fn print_result(result: io::Result<FileId>) {
     match result {
         Ok(file_id) => println!("{file_id:?}"),
-        Err(error) => println!("Error: {error}"),
+        Err(error) => eprintln!("Error: {error}"),
     }
 }
--- a/file-id/src/lib.rs
+++ b/file-id/src/lib.rs
@@ -85,21 +85,21 @@ pub enum FileId {
 
 impl FileId {
     pub fn new_inode(device_id: u64, inode_number: u64) -> Self {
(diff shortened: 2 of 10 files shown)


file-id/src/lib.rs
```diff
@@ -85,21 +85,21 @@ pub enum FileId {
 
 impl FileId {
     pub fn new_inode(device_id: u64, inode_number: u64) -> Self {
-        FileId::Inode {
+        Self::Inode {
```
0xpr03, line 88: I think replacing the Name with Self is only reducing readability and doesn't really remove any duplication (FileId -> Self) when you use also `use FileId::*;` and drop the prefix completely. Either way I think it's fine as it is.

file-id/bin/file_id.rs
```diff
@@ -22,6 +22,6 @@ fn print_file_id(path: &str) {
 fn print_result(result: io::Result<FileId>) {
     match result {
         Ok(file_id) => println!("{file_id:?}"),
-        Err(error) => println!("Error: {error}"),
+        Err(error) => eprintln!("Error: {error}"),
```
0xpr03, line 25: I mean, you can do that but this is just an example, so it won't matter at all. (Also there is an unrelated import change in that commit.)

0xpr03 (changes requested):
I would leave the `Self` parts for now, unless you want to use these changes to get towards a bigger feature. Adding only changes for subjective code readability is sadly adding git log/blame noise.

coswat:
oh ok

PR #540, opened 2023-10-07
Fixed the Event of watch dir Deletion

### LINKED ISSUE #493: Delete a directory which is watched returns Event { kind: Remove(File) }
#### System details

- OS/Platform name and version: `Fedora 38 `
- Rust version (if building from source): `rustc 1.70.0 (90c541806 2023-05-31)`
- Notify version (or commit hash if building from git): `6.0.0`
- On Linux: Kernel version: `6.3.6-200.fc38.x86_64`

1. call `watcher.watch(path, non_recursive)` on a directory
2. `rm -r path`

#### What you expected
Receive `Event { kind: Remove(Folder) }`

#### What happened
Receive `Event { kind: Remove(File) }`

zeroishero:
Resolves #493
Deleting watched dir only returns Delete self and Ignore event. Unlike deleting other folders, it doesn't return events to check whether it is dir or not. 

I stored whether path is dir or not in watches. Taking metadata after delete event triggers is not possible.

### DIFF
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -37,6 +37,7 @@ struct EventLoop {
     inotify: Option<Inotify>,
     event_handler: Box<dyn EventHandler>,
     // PathBuf -> (WatchDescriptor, WatchMask, is_recursive, is_dir)
+    /// PathBuf -> (WatchDescriptor, WatchMask, is_recursive, is_dir)
     watches: HashMap<PathBuf, (WatchDescriptor, WatchMask, bool, bool)>,
     paths: HashMap<WatchDescriptor, PathBuf>,
     rename_event: Option<Event>,
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -306,13 +306,16 @@ impl EventLoop {
                                             None => RemoveKind::Other,
                                         }
                                     }
-                                    None => RemoveKind::Other,
+                                    None => {
+                                        log::trace!("No patch for DELETE_SELF event, may be a bug?");
+                                        RemoveKind::Other
+                                    },
                                 };
                                 evs.push(
                                     Event::new(EventKind::Remove(remove_kind))
                                         .add_some_path(path.clone()),
                                 );
-                                remove_watch_by_event(&path, &self.watches, &mut remove_watches)
+                                remove_watch_by_event(&path, &self.watches, &mut remove_watches);
                             }
                             if event.mask.contains(EventMask::MODIFY) {
                                 evs.push(
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -297,9 +297,9 @@ impl EventLoop {
                                 remove_watch_by_event(&path, &self.watches, &mut remove_watches);
                             }
                             if event.mask.contains(EventMask::DELETE_SELF) {
-                                let remove_kind = match path.clone() {
+                                let remove_kind = match &path {
                                     Some(watched_path) => {
-                                        let current_watch = self.watches.get(&watched_path);
+                                        let current_watch = self.watches.get(watched_path);
                                         match current_watch {
                                             Some(&(_, _, _, true)) => RemoveKind::Folder,
                                             Some(&(_, _, _, false)) => RemoveKind::File,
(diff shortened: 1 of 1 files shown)


notify/src/inotify.rs
```diff
@@ -35,7 +37,7 @@ struct EventLoop {
     event_loop_rx: Receiver<EventLoopMsg>,
     inotify: Option<Inotify>,
     event_handler: Box<dyn EventHandler>,
-    watches: HashMap<PathBuf, (WatchDescriptor, WatchMask, bool)>,
+    watches: HashMap<PathBuf, (WatchDescriptor, WatchMask, bool, bool)>,
```
0xpr03, line 40:
```suggestion
    /// - watches HashMap Value tuple (WatchDescriptor, WatchMask, is_recursive, is_dir)
    watches: HashMap<PathBuf, (WatchDescriptor, WatchMask, bool, bool)>,
```

notify/src/inotify.rs
```diff
@@ -296,6 +296,28 @@ impl EventLoop {
                                 );
                                 remove_watch_by_event(&path, &self.watches, &mut remove_watches);
                             }
+                            if event.mask.contains(EventMask::DELETE_SELF) {
+                                evs.push(
+                                    Event::new(EventKind::Remove(match path.clone() {
```
0xpr03, line 301: Can you refactor this so there is no hidden match in the value creation ? Not sure but rustfmt might also move this around a fair bit..

notify/src/inotify.rs
```diff
@@ -296,6 +296,28 @@ impl EventLoop {
...
+                                        Some(path) => {
+                                            let current_watch = self.watches.get(&path);
+                                            match current_watch {
+                                                Some(&(_, _, _, is_dir)) => {
+                                                    if is_dir {
```
0xpr03, line 306: this could also be collapsed into a single match over the cases Some(..,true,), Some(..,false), None

0xpr03 (changes requested):
Thanks! I've added some comments, let me know if you want to leave them to me or refactor those parts by yourself.

notify/src/inotify.rs
zeroishero: Done

notify/src/inotify.rs
zeroishero: I took the logic to check for Dir Type outside push method.

zeroishero:
I added the comment for watches HashMap.

notify/src/inotify.rs
zeroishero: done

0xpr03: can this be a doc comment ? `///`

0xpr03: thanks, but please also remove the upper one :)

zeroishero:
@0xpr03 is there anything I should add?

0xpr03:
Any reason to close it ? I did ask for another thing: The unwrap you're using seems like something that could crash. Otherwise the PR is fine.

zeroishero:
Oh.. I thought that didn't passed review and closed it. I didn't see the new change. I will check the unwrap one.

zeroishero:
@0xpr03 I used unwrap because the if clause checks for is_none().  Can it crash?

anyway, I refactored it to not use unwrap.

notify/src/inotify.rs
zeroishero: Done. I thought you meant the comment above.

zeroishero:
@0xpr03 are there more changes I should do before it gets approved?

0xpr03:
Looks good, I'm just too busy right now and don't wanna rush things.

0xpr03 (approved):

0xpr03:
Huh so apparently github never send my review.. Sorry for that..

notify/src/inotify.rs
```diff
@@ -296,6 +297,26 @@ impl EventLoop {
                                 );
                                 remove_watch_by_event(&path, &self.watches, &mut remove_watches);
                             }
+                            if event.mask.contains(EventMask::DELETE_SELF) {
+                                let remove_kind: RemoveKind;
+
+                                if path.is_none() {
+                                    remove_kind = RemoveKind::Other
+                                } else {
+                                    let watched_path = path.clone().unwrap();
```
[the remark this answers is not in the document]
hexofyore: considering it's in else branch which checks for is_none condition, it shouldn't fail on unwrap. 
The clone is there coz it needs to be owned by Event but path is in notify.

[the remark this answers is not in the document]
zeroishero: considering it's in else branch which checks for is_none condition, it shouldn't fail on unwrap.
The clone is there coz it needs to be owned by Event but path is in notify.

PR #543, opened 2023-10-30
improve first example in the docs

mumbleskates:
The first example is misleading in that it doesn't show code that will produce a useful example, but instead exits instantly with no standing example of how to usefully get events to the main thread.

### DIFF
--- a/notify/src/lib.rs
+++ b/notify/src/lib.rs
@@ -102,16 +102,17 @@
 //!
 //! ```rust
 //! # use std::path::Path;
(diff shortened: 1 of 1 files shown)


0xpr03:
this is still blocked by CI failing

mumbleskates:
Thanks, should be unblocked now; i removed the duplicate import it was failing on.

JohnTitor (approved):

JohnTitor:
Thanks!

PR #547, opened 2023-11-13
Update inotify to 0.10

fornwall:
See the [CHANGELOG](https://github.com/hannobraun/inotify-rs/blob/main/CHANGELOG.md).

### DIFF
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -465,10 +465,10 @@ impl EventLoop {
             None => return Err(Error::watch_not_found().add_path(path)),
             Some((w, _, is_recursive, _)) => {
                 if let Some(ref mut inotify) = self.inotify {
+                    let mut inotify_watches = inotify.watches();
                     log::trace!("removing inotify watch: {}", path.display());
 
-                    inotify
-                        .watches()
+                    inotify_watches
                         .remove(w.clone())
                         .map_err(|e| Error::io(e).add_path(path.clone()))?;
                     self.paths.remove(&w);
@@ -477,8 +477,7 @@ impl EventLoop {
                         let mut remove_list = Vec::new();
                         for (w, p) in &self.paths {
                             if p.starts_with(&path) {
-                                inotify
-                                    .watches()
+                                inotify_watches
                                     .remove(w.clone())
                                     .map_err(|e| Error::io(e).add_path(p.into()))?;
                                 self.watches.remove(p);
@@ -497,9 +496,9 @@ impl EventLoop {
 
     fn remove_all_watches(&mut self) -> Result<()> {
         if let Some(ref mut inotify) = self.inotify {
+            let mut inotify_watches = inotify.watches();
             for (w, p) in &self.paths {
-                inotify
-                    .watches()
+                inotify_watches
                     .remove(w.clone())
                     .map_err(|e| Error::io(e).add_path(p.into()))?;
             }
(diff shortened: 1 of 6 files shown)


fornwall:
Hi @0xpr03, could you have a look at this PR?

notify/src/inotify.rs
```diff
@@ -307,9 +307,11 @@ impl EventLoop {
                                         }
                                     }
                                     None => {
-                                        log::trace!("No patch for DELETE_SELF event, may be a bug?");
+                                        log::trace!(
```
0xpr03, line 310: was this part of cargo fmt ?

0xpr03:
Is there a strong reason to push for 0.10 ASAP ? I don't see any real reasons in their changelog.

notify/src/inotify.rs
```diff
@@ -471,7 +474,8 @@ impl EventLoop {
                         for (w, p) in &self.paths {
                             if p.starts_with(&path) {
                                 inotify
-                                    .rm_watch(w.clone())
+                                    .watches()
```
0xpr03, line 477: The watches() call seems to perform an Arc clone and creating some sort of Struct containing a [FdGuard](https://github.com/hannobraun/inotify-rs/blob/39b64297a6d3dd2706652c0e19d061287a4e04be/src/fd_guard.rs#L22). It might be in our interest to use one watches() call for at least the whole loop.

0xpr03 (changes requested):
Blocked by #549.
And we might want to look into re-using watches() for hot loops.

0xpr03:
Thanks, I've added one comment regarding the watches() overhead.

fornwall:
> Is there a strong reason to push for 0.10 ASAP ? I don't see any real reasons in their changelog.

The main driver here is to avoid duplication - other crates in the ecosystem are using 0.10 (for instance https://gitlab.com/gilrs-project/gilrs), and there is a desire (or rule) in many projects to avoid pulling in multiple versions of the same crate.

So in general, unless there is a good reason, I think the latest released version of a crate should be used.

notify/src/inotify.rs
fornwall: Yes, it was.

notify/src/inotify.rs
fornwall: Thanks, will look into this.

fornwall:
Tested this a bit, and it seems we need to adopt to: https://github.com/hannobraun/inotify-rs/pull/190

notify/src/inotify.rs
fornwall: Done in b70b72b79f9ad273a2f36a3d435a97f069b360ca

fornwall:
> Thanks, I've added one comment regarding the watches() overhead.

Thank you for the review! Addressed that one in b70b72b79f9ad273a2f36a3d435a97f069b360ca

MarijnS95:
> The main driver here is to avoid duplication - other crates in the ecosystem are using 0.10 (for instance https://gitlab.com/gilrs-project/gilrs), and there is a desire (or rule) in many projects to avoid pulling in multiple versions of the same crate.

Yeah, same for us, we'd love to see both crates in sync :)

0xpr03 (approved):

0xpr03:
Thanks, I got to merge the blocker and fixed one small readme formatting error while rebasing for (valid) CI results.

0xpr03:
Now let's see, there is some hope I can get around to a new release this weekend, but no promises.

PR #551, opened 2023-12-06
Add fail on permissions flag

amircodota:
I ran into a case where, specifically in inotify, if I'm watching a directory and the user does not have permissions on a subdirectory, the watch will fail.

Added a flag in the Config struct to change this behavior (still the default).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 tests/last-fails
 tests/last-run.log
 .cargo
+.idea/
+*.iml
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -10,13 +10,13 @@ use crate::{bounded, unbounded, BoundSender, Receiver, Sender};
 use inotify as inotify_sys;
 use inotify_sys::{EventMask, Inotify, WatchDescriptor, WatchMask};
 use std::collections::HashMap;
-use std::env;
 use std::ffi::OsStr;
 use std::fs::metadata;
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@
 tests/last-fails
 tests/last-run.log
 .cargo
+.idea/
+*.iml
```
JohnTitor, line 8: Could you add them to your global gitignore instead? Accepting any IDEs makes our gitignore fat and hard to maintain.

notify/src/inotify.rs
```diff
@@ -90,7 +91,11 @@ fn remove_watch_by_event(
 }
 
 impl EventLoop {
-    pub fn new(inotify: Inotify, event_handler: Box<dyn EventHandler>) -> Result<Self> {
+    pub fn new(
+        inotify: Inotify,
+        event_handler: Box<dyn EventHandler>,
+        fail_on_no_permissions: bool,
```
JohnTitor, line 97: I'd prefer passing a config like this via a wrapper (a struct) instead, otherwise args could be too long and we have to expect breaking changes each time we modify.

0xpr03: yep, please pass the config down if possible

0xpr03 (changes requested):
thanks for the PR - if there a reason not to fail if we're lacking permissions, making this a breaking change instead for the next major version ?

PR #556, opened 2024-01-19
[notify-debouncer-full] mark as compatible with rstest 0.18.x

michel-slm:
We are packaging this in Fedora, and tests pass using rstest 0.18.x that is in our repositories

See https://bugzilla.redhat.com/show_bug.cgi?id=2257209


JohnTitor:
I'd just declare `rstest = "0.18"`, I don't think the flexibility is useful on the dev deps.

michel-slm:
> I'd just declare `rstest = "0.18"`, I don't think the flexibility is useful on the dev deps.

good point! updated the commit

JohnTitor (approved):
Thank you!

PR #557, opened 2024-01-21
Manage file ID cache paths in debouncer-full

dfaust:
- CHANGE: Manage root folder paths for the file ID cache automatically. **breaking**

  This simplifies the usage of the `debouncer-full`. By default it will now maintain the file id cache on its own.
  It should make it impossible to misuse the feature.

- CHANGE: Add `RecommendedCache`, which automatically enables the file ID cache on Windows and MacOS
  and disables it on Linux, where it is not needed.

  inotify emits rename cookies, so the file id cache can be disabled on Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,26 @@
 # Changelog
 
-v5 maintenance branch is on `v5_maintenance` after `5.2.0`  
+v5 maintenance branch is on `v5_maintenance` after `5.2.0`
+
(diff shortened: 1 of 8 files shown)


ForsakenHarmony:
When is this getting released?

dfaust:
> When is this getting released?

@0xpr03 I just pushed a bunch of PRs. When they are merged, we should release a new version.

0xpr03:
Will take a look tomorrow.

PR #559, opened 2024-01-25
Move event types into separate crate

### LINKED ISSUE #487: Proposal: split Event structure into its own crate (notify-events?)
This is a proposal to only split out the `notify/src/event.rs` file into its own crate, which would have no required dependencies, and serde only when the feature is enabled.

The `Event` struct and child enums are pretty infrequently changed, and it would be helpful to downstream crates/users to be able to use the notify event types without pulling the entire notify runtime.

My specific motivation is for a usage across an IPC or network barrier, where one side does the file watching and the other wants to deserialize into the proper types.

dfaust:
Move event types into separate crate.

Closes #487.


0xpr03 (approved):
Looks good. Thanks for the work!

Let me know if you want this merged. I'm always unsure whether PRs are "WIP" / informative or actually "ready for merge" when they're opened without any flaws.

dfaust:
Please merge it.
I'll try to be clear in case I have a WIP PR.

0xpr03:
I'll do a crate push  / release once I'm back home today.

0xpr03:
>  I'll try to be clear in case I have a WIP PR.

Sounds good. I've just seen a lot of PRs where in the time it took me to actually review them, they got additional commits for fixups. So I would've prematurely merged stuff if I would've been faster. (Finding typos is something I always do afterwards..)

dfaust:
That's true. Happens to me all the time. And if I can, I fix those issues immediately. But if you merge my PR "too quick", it's still my fault for requesting bad code to be merged. So don't worry - and it can always be fixed later.

dfaust:
> I'll do a crate push / release once I'm back home today.

Maybe hold back on that and check #560.

PR #568, opened 2024-02-21
Add `serialization-compat-6` feature

dfaust:
This feature restores the serialization behavior of notify 6.

CC: @passcod 

I screwed up a bit and this PR is based on my previous on (#567).


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-
+resolver = "2"
```
0xpr03, line 2: does that increase the MSRV ?

PR #569, opened 2024-02-21
MSRV 1.72

dfaust:
- Raise MSRV to 1.72
- Remove internal use of crossbeam because `std::sync::mpsc::Sender` implements `Sync` since version 1.72, so there is no need for crossbeam anymore
- Specify dependencies in workspace

I screwed up a bit and this PR is based on my previous on (#567).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - 1.63.0 # MSRV
+          - 1.72.1 # MSRV
           - stable
           - nightly
         os:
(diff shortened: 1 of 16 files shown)


0xpr03:
Let's merge this with #568 as 7.0 and try to get everything else into 6.X

Or is there anything else that would be major breaking ?

0xpr03:
Ah yeah, my bad. Nevermind then, break all the things.

0xpr03:
This also needs a rebase

0xpr03:
I'll re-rebase this when I'm back at home. Then I wall have to look at the outstanding issues and PR that accumulated over the past ~2 months and push the release finally over the line..

Thanks for all the work!

dfaust:
Can we merge this? I already based my next PR on this one.

PR #573, opened 2024-03-08
Support watching the same inode through multiple paths.

### LINKED ISSUE #572: inotify: files pointed at by multiples symlinks generate events only for one path
#### System details

- OS/Platform name and version: Linux 6.5.0 - Ubuntu 22.04.3
- Rust version (if building from source): `rustc --version`: `rustc 1.75.0-nightly (0f44eb32f 2023-11-09)`
- Notify version (or commit hash if building from git): 6.1.1

- If you're coming from a project that makes use of Notify, what it is, and a link to the downstream issue if there is one: buck2 (https://buck2.build)
- Filesystem type and options: ext4
- On Linux: Kernel version: 6.5.0

I have a directory structure similar to the following:
```
- root
| - common
| | - file.json
| - tests
| | - test_1
| | | - common -> ../../common
```

We put a recursive watch on `root`, and then modify `root/common/file.json`.

#### What you expected
Either one of:
* Notify events for both `root/common/file.json` and `root/tests/test_1/common/file.json`.
* Notify events for just `root/common/file.json`

#### What happened
Only a notify event is fired for `root/tests/test_1/common/file.json`.

hugwijst:
Support watching the same inode through multiple paths.

As a single change event can now refer to multiple paths, use "path groups" for renames to differentiate sources and destinations.

Fixes #572.

### DIFF
--- a/notify-types/src/event.rs
+++ b/notify-types/src/event.rs
@@ -309,8 +309,13 @@ pub struct Event {
     /// creation, they should all go in this `Vec`. Otherwise, using the `Tracker` attr may be more
     /// appropriate.
     ///
-    /// The order of the paths is likely to be significant! For example, renames where both ends of
-    /// the name change are known will have the "source" path first, and the "target" path last.
+    /// The order of the paths is likely to be significant! Furthermore, the paths might be part of
+    /// different "groups". For example, renames where both ends of the name change are known will
+    /// have the "source" paths first, followed by the "target" paths. These two sets of paths are
+    /// in distinct groups.
+    ///
+    /// If there are different path groups, the `path_group_split_index` attribute indicates the
+    /// first index with paths in the second group.
     pub paths: Vec<PathBuf>,
 
     // "What should be in the struct" and "what can go in the attrs" is an interesting question.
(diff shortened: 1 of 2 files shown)


notify-types/src/event.rs
```diff
@@ -309,8 +309,13 @@ pub struct Event {
...
-    /// The order of the paths is likely to be significant! For example, renames where both ends of
-    /// the name change are known will have the "source" path first, and the "target" path last.
+    /// The order of the paths is likely to be significant! Furthermore, the paths might be part of
+    /// different "groups". For example, renames where both ends of the name change are known will
+    /// have the "source" paths first, followed by the "target" paths. These two sets of paths are
+    /// in distinct groups.
+    ///
+    /// If there are different path groups, the `path_group_split_index` attribute indicates the
+    /// first index with paths in the second group.
     pub paths: Vec<PathBuf>,
```
hugwijst, line 319: As this is a behavioral change, we probably want some kind of API change so users explicitly decide how to handle multiple paths in one group, or this newly introduced concept of 2 different groups.

I didn't want to break this into two separate vectors, as to not increase the size of the `Event` struct. Maybe we should make `paths` private and have a `fn path_groups(&self) -> (&[PathBuf], Option<&[PathBuf]>)` method?

hugwijst:
Ping @dfaust, @0xpr03. I'm wondering if there is a less intrusive solution?

hugwijst:
Ping @passcod, @dfaust, @0xpr03, would any of you be able to review?

The current implementation is buggy if the watched directory contains two hard or symbolic links to the same file. I can see an argument for symbolic links to not be followed at all, but something like this change is needed for hard links I think.

notify-types/src/event.rs
passcod: How about instead of having the api on the attrs interact directly with the fields, there's a more intuitive API on Event like `renamed_from_paths() -> Option<&[Path]>` and `renamed_to_paths() -> Option<&[Path]>` that interprets the kind/paths/attr information?

hugwijst:
After revisiting this PR I think the ignoring symlink while walking the directories is the better solution to my problem: #635. Some of the changes here might still be useful to handle hardlinks.

notify-types/src/event.rs
passcod: I was, yes.

PR #582, opened 2024-04-04
Update debouncer_full.rs - compatible with 0.3.1

TheCaptainMoo:
Updated some of the imports and start-up of the watcher to be compatible with debouncer_full version 0.3.1

Some of the code for the example seemed to be outdated and has been modified slightly to work with the updated debouncer_full version 0.3.1


dfaust:
Thanks for your PR, but I'm going to close it because the `watcher` method has been deprecated.

PR #584, opened 2024-04-14
Update `windows-sys` to `0.52`

skyfloogle:
Required for building on certain targets on newer versions of Rust (see https://github.com/microsoft/windows-rs/pull/2774).


waywardmonkeys:
This is done now on `main` in other commits.

skyfloogle:
Thanks!

PR #589, opened 2024-05-02
Update README.md to include Deno

garikAsplund:
Added Deno since it's a huge project. Also reordered list to be alphabetical--was just one off.


0xpr03:
Thanks!

PR #590, opened 2024-05-06
Fix blender save

dfaust:
- FIX: ordering of debounced events when multiple files are modified and renamed (eg. during a safe save performed by Blender)

The debouncer is getting increasingly complicated. It took me a few tries to design this change as defensively as possible. I hope it doesn't break anything.

This PR is based on #569.

### DIFF
--- a/notify-debouncer-full/CHANGELOG.md
+++ b/notify-debouncer-full/CHANGELOG.md
@@ -3,8 +3,10 @@
 ## unreleased
 
 - CHANGE: emit `remove` events even if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
+- FIX: ordering of debounced events when multiple files are modified and renamed (eg. during a safe save performed by Blender) [#590]
 
 [#900]: https://github.com/notify-rs/notify/issues/900
+[#590]: https://github.com/notify-rs/notify/issues/590
 
 ## debouncer-full 0.8.0-rc.1 (2026-04-16)
 
--- a/notify-debouncer-full/src/lib.rs
+++ b/notify-debouncer-full/src/lib.rs
@@ -61,6 +61,7 @@
 //!
 //! As all file events are sourced from notify, the [known problems](https://docs.rs/notify/latest/notify/#known-problems) section applies here too.
 
+mod block_manager;
 mod cache;
 mod time;
 
@@ -71,8 +72,7 @@ mod testing;
 mod file_id_map;
 
 use std::{
-    cmp::Reverse,
-    collections::{BinaryHeap, VecDeque},
+    collections::{HashMap, VecDeque},
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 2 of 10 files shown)


notify-debouncer-full/src/lib.rs
```diff
@@ -209,48 +255,62 @@ impl<T: FileIdCache> DebounceDataInner<T> {
...
-            }
+                kind_index.insert(event.kind, events_expired.len());
 
-            if !queue.events.is_empty() {
-                queues_remaining.insert(path, queue);
+                self.blocking.remove_blocker(&path, event.time);
+
+                events_expired.push(event);
+            } else {
+                self.queues.get_mut(&path).unwrap().events.push_front(event);
```
0xpr03, line 303: same here for unwrap stuff - want to avoid crashing and rather report an error

0xpr03 (commented):
I think we will need way more documentation to keep this usable in the future

notify-debouncer-full/src/lib.rs
```diff
@@ -455,7 +503,12 @@ impl<T: FileIdCache> DebounceDataInner<T> {
         }
 
         if let Some(target_queue) = self.queues.get_mut(&event.paths[0]) {
-            if !target_queue.was_created() {
+            if target_queue.was_removed() {
+                let event = target_queue.events.pop_front().unwrap();
```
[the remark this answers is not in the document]
dfaust: Yes, unwrap is safe here, I added a comment to explain.
Maybe we can refactor the code to make it more elegant and remove the unwrap, but the function doesn't return a Result, because it can only fail if there's an implementation error.

notify-debouncer-full/src/lib.rs
dfaust: Same as above.

dfaust:
> I think we will need way more documentation to keep this usable in the future

Agreed. I will add more comments as I work on it.

0xpr03:
We can merge this for now, if you want. I just would like to document this, because I fear we will otherwise have a ton of regressions in the future.

dfaust:
Then let's not merge it for now. I will document it better and try to fix the original issue.

While the PR does improve the behavior of the debouncer, it doesn't actually fix the blender issue.

notify-debouncer-full/CHANGELOG.md
```diff
@@ -3,8 +3,10 @@
 ## unreleased
 
 - CHANGE: emit `remove` events even if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
+- FIX: ordering of debounced events when multiple files are modified and renamed (eg. during a safe save performed by Blender) [#590]
```
peddermaster2, line 6: Should `safe save` be just `save`?

```suggestion
- FIX: ordering of debounced events when multiple files are modified and renamed (eg. during a save performed by Blender) [#590]
```

dfaust: Nope. I'm specifically talking about the "safe" saving process where a new file is written first and then moved to the original location.

PR #592, opened 2024-05-12
Fix build for DragonFlyBSD.

waywardmonkeys:
The correct `target_os` for this is `dragonfly`, not `dragonflybsd`.

This was introduced in 8399e4195b31f6f188109363292ed220226146f4.


waywardmonkeys:
This can be seen by doing a build with `nightly` after the `check-cfg` feature was added.

0xpr03:
Fix is valid as documented per https://doc.rust-lang.org/reference/conditional-compilation.html#target_os

Thanks!

PR #593, opened 2024-05-12
Fix typos.

waywardmonkeys:


notify/src/windows.rs
```diff
@@ -138,8 +138,8 @@ impl ReadDirectoryChangesServer {
 
             unsafe {
                 // wait with alertable flag so that the completion routine fires
-                let waitres = WaitForSingleObjectEx(self.wakeup_sem, 100, 1);
-                if waitres == WAIT_OBJECT_0 {
+                let waiters = WaitForSingleObjectEx(self.wakeup_sem, 100, 1);
```
0xpr03, line 141: that's intentional

0xpr03 (changes requested):
Looks good apart from the two lines

notify/src/windows.rs
waywardmonkeys: Whoops, I'd meant to remove those. Sorry about that! Removed these now.

0xpr03:
Thanks!

Issue #596, opened 2024-05-19
Using kqueue on mac can result in "Too many open files (os error 24)"


alanwsmith:
#### System details

- OS/Platform name and version: 

    Apple M1 Max - macOs Sonoma 14.4.1

- Rust version (if building from source): `rustc --version`: 

    rustc 1.79.0-nightly (1cec373f6 2024-04-16)

- Notify version (or commit hash if building from git): 

    notify = "6.1.1"

#### What you did

I tried using the `macos_kqueue` feature via this in my Cargo.toml file:

```
notify = { version = "6.1.1", default-features = false, features = ["macos_kqueue"] }
```

I then set up the watcher to watch 300 files. (see code sample at the end of this post)

#### What you expected

I expected notify to watch the files and trigger when they changed

#### What happened

Instead, I got this error as soon as I started the process

```
Error: Error { kind: Generic("Too many open files (os error 24) about [\"/Users/alan/Desktop/notify-test/./test_files/268.txt\"]"), paths: [] }
```

#### Notes

- This is the script I'm using to run the tests:

```toml
[dependencies]
notify = { version = "6.1.1", default-features = false, features = ["macos_kqueue"] }
```

```rust
use notify::{Watcher, RecursiveMode, Result};
use std::path::PathBuf;
use std::fs;

fn main() -> Result<()> {
    let test_dir = PathBuf::from("./test_files");
    fs::create_dir_all(&test_dir).unwrap();

    for f in 0..300 {
        let file_path = test_dir.clone().join(format!("{}.txt", f));
        fs::write(file_path, "x").unwrap();
    }

    let mut watcher = notify::recommended_watcher(|res| {
        match res {
           Ok(event) => println!("event: {:?}", event),
           Err(e) => println!("watch error: {:?}", e),
        }
    })?;

    watcher.watch(&test_dir, RecursiveMode::Recursive)?;
    Ok(())
}
```

- I found [this issue](https://stackoverflow.com/a/56843690/102401) which talks about the max number of allowed files imposed by the os. My default from `ulimit -n` was 256. If I change the limit to 500 with `ulimit -n 500` the process works fine. If I then change it back with `ulimit -n 256` it fails again with the same error

[...]

PR #597, opened 2024-05-19
Changed tokio crossbeam example from kqueue to fsevent

alanwsmith:
- This pull request changes the example code for turning off crossbeam for tokio to use `macos_fsevent` instead of `macos_kqueue`

- I'm suggesting this change based off [the issue I'm seeing with too many open files with kqueue](https://github.com/notify-rs/notify/issues/596)

- My thinking is that this will help prevent folks running into the `kqueue` by using `fsevent` instead

### DIFF
--- a/notify/src/lib.rs
+++ b/notify/src/lib.rs
@@ -36,10 +36,9 @@
 //! You can disable crossbeam-channel, letting notify fallback to std channels via
 //!
 //! ```toml
-//! notify = { version = "6.1.1", default-features = false, features = ["macos_kqueue"] }
-//! // Alternatively macos_fsevent instead of macos_kqueue
+//! notify = { version = "6.1.1", default-features = false, features = ["macos_fsevent"] }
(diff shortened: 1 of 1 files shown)


notify/src/lib.rs
```diff
@@ -36,10 +36,9 @@
 //! You can disable crossbeam-channel, letting notify fallback to std channels via
 //!
 //! ```toml
-//! notify = { version = "6.1.1", default-features = false, features = ["macos_kqueue"] }
-//! // Alternatively macos_fsevent instead of macos_kqueue
```
0xpr03, line 40: please don't remove this, if you only watch a little amount of files, [kqueue is totally valid](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html)

0xpr03 (changes requested):

PR #602, opened 2024-06-04
Fix UB in `windows::start_read`

encounter:
The primary issue is that `mem::transmute` from `isize` to `Box<_>` (without first casting to `*mut _`) is undefined behavior.

[Rust Playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8babb8f7b34293b67bc06302b9d606c9) (check with Tools -> Miri)

On Rust v1.78.0+, this ends up crashing with `STATUS_ILLEGAL_INSTRUCTION` when `ReadDirectoryChangesW` fails and this branch is hit in release mode. One easy way to hit this branch is to attempt to create a notifier for `\\wsl.localhost\...` or similar.

While this could be fixed by simply adding `as *mut ReadDirectoryRequest`, this cleans up the overall unsafe logic to be more readable and idiomatic Rust.

### DIFF
--- a/notify/src/windows.rs
+++ b/notify/src/windows.rs
@@ -8,10 +8,10 @@
 use crate::{bounded, unbounded, BoundSender, Config, Receiver, Sender};
 use crate::{event::*, WatcherKind};
 use crate::{Error, EventHandler, RecursiveMode, Result, Watcher};
+use std::alloc;
 use std::collections::HashMap;
 use std::env;
 use std::ffi::OsString;
-use std::mem;
 use std::os::raw::c_void;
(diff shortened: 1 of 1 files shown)


0xpr03:
Ugh looks like freebsd needs a CI fix. Let's see when I can actually fix this all up.

0xpr03:
Hm I would have liked #604 to be a separate commit. This way it is harder to diff. Whole thing is already hairy enough.

0xpr03:
I will merge the other one first, that will also credit the author and merge part we know to be working. Please do me the favor and rebase on top of that.

PR #603, opened 2024-06-11
Add missing winapi features to fix #601

### LINKED ISSUE #601: v4 of Notify lacks `std` and `winerror` features of `winapi`
#### System details

- OS/Platform name and version: Windows 11 23H2
- Rust version (if building from source): `rustc --version`: rustc 1.78.0
- Notify version (or commit hash if building from git): v4.0.X

No other details are relevant as this is a build-time issue.

In a word: nothing. Users started reporting that using `cargo install` for a project I maintain no longer worked, so I investigated it. This has lead me here.

To be explicit: if you attempt to compile a project that is using notify v4 without a lock file, it will currently fail outright. This includes projects like `cargo-watch`, which use v4 of Notify.

#### What you expected

I expect that projects using notify v4 will compile, ideally.

#### What happened

Notify version 4 will fail to compile because it lacks the `std` and `winerror` features of `winapi`. Up until relatively recently, these features were enabled by a downstream dependency named [`winapi-util`](https://github.com/BurntSushi/winapi-util). They no longer are, which has caused the missing features to become obvious.

MikeOnTea:
Untested, but hopefully fixes #601.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ fsevent = "0.4"
 fsevent-sys = "2"
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.8", features = ["fileapi", "handleapi", "ioapiset", "minwinbase", "synchapi", "winbase", "winnt"] }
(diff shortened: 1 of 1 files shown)


0xpr03:
Verified by manual testing too

PR #604, opened 2024-06-21
fix UD behavior on windows - causes a crash on rust 1.78

amircodota:
Fixes the issue described [here](https://github.com/rust-lang/rust/issues/126195)


encounter:
Just FYI, this is also fixed by https://github.com/notify-rs/notify/pull/602

PR #608, opened 2024-06-27
Update `rstest`

waywardmonkeys:
Remove the dependency on `rstest` for `notify` as it wasn't used.

Update the other dependencies on `rstest`.


0xpr03:
Thanks, can we unify this by using a workspace dev crate ?

waywardmonkeys:
@0xpr03 Rebased and updated!

PR #610, opened 2024-06-30
Rebase of #569

0xpr03:
Rebuild #569 on current master


0xpr03:
cc @dfaust as a sanity check

dfaust:
@0xpr03: Looks reasonable!

Issue #110, opened 2017-01-12
When trying to watch non-existing file or directory, report a single error kind instead of the native one


dfaust:
Linux: `Io(Error { repr: Os { code: 2, message: "No such file or directory" } })`
Windows: `Generic("Input watch path is neither a file nor a directory.")`
macOS: `PathNotFound`

Any opinions?

passcod:
This will be handled above the backend in v5.

However, it will still be a problem in race conditions when the file disappears between the time the frontend checks for it and the backend attempts to watch it. That will need to be fixed somehow, but I haven't yet figured out what path I want to take for it.

passcod:
This is solved in v5 as file-does-not-exist errors are a specific variant now.

The race condition problem is also solved as both a core and a backend concern. Basically because of #64 and how that's implemented we retry creating backends if it fails on specific paths, without those specific paths.

Closing before v5 release to clear things up a bit.

PR #611, opened 2024-07-01
Make `ErrorKind::PathNotFound` consistent on inotify

y5c4l3:
Before this patch, watching a nonexistent path using inotify backend would give an `ErrorKind::Io` variant. This behavior was inconsistent with other backends.

I am not sure if this is expected but taking #110 into consideration I think the path error ought to be consistent.

Run the inotify watcher on Linux:
```rust
let error = watcher.watch(Path::new("/nonexistent"), notify::RecursiveMode::Recursive).unwrap_err();
```

Before:
```
Error { kind: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" }), paths: [] }
```

After:
```
Error { kind: PathNotFound, paths: [] }
```

### DIFF
--- a/notify/src/error.rs
+++ b/notify/src/error.rs
@@ -81,6 +81,15 @@ impl Error {
         Self::new(ErrorKind::Io(err))
     }
 
+    /// Similar to [`Error::io`], but specifically handles [`io::ErrorKind::NotFound`].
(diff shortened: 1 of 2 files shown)


0xpr03:
Thanks!

PR #612, opened 2024-07-02
Set `WatchMask::OPEN` on inotify

y5c4l3:
This fixes the problem of `Access(Open(_))` variant not being reported.

The event variant is constructed, but `watchmask` does not have `WatchMask::OPEN` set.

https://github.com/notify-rs/notify/blob/f9c15ebf2cfd6a45f57c45290d0579846739fdbd/notify/src/inotify.rs#L354-L361

### DIFF
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -417,6 +417,7 @@ impl EventLoop {
     ) -> Result<()> {
         let mut watchmask = WatchMask::ATTRIB
             | WatchMask::CREATE
+            | WatchMask::OPEN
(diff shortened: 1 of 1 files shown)


0xpr03:
Can you mention which issue this is referring to ?

y5c4l3:
> Can you mention which issue this is referring to ?

@0xpr03 I don't think there is an issue for the problem. I encountered it and it was a quick fix, so I directly made a pull request.

0xpr03:
Thanks!

PR #600, opened 2024-05-21
Minor: Fixed lint warnings seen while running "cargo doc".

martinfrances107:
Just fixing some lint warnings seen while running cargo doc

I hope this helps.

PR #615, opened 2024-07-14
file-id: Fix rustdoc `bare-url` lints

waywardmonkeys:


waywardmonkeys:
This was also done in #600.

PR #618, opened 2024-07-14
Fix: Low Res and high Res ID inversion

CalunVier:
It looks like the get_low_res_file_id() and get_high_res_file_id() functions in the file_id crate are mistakenly calling opposite underlying functions. At least that's how I understand it, there's something wrong here.

### DIFF
--- a/file-id/src/lib.rs
+++ b/file-id/src/lib.rs
@@ -129,7 +129,7 @@ pub fn get_file_id(path: impl AsRef<Path>) -> io::Result<FileId> {
 pub fn get_low_res_file_id(path: impl AsRef<Path>) -> io::Result<FileId> {
     let file = open_file(path)?;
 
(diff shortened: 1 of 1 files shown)


CalunVier:
Github Action CI has pass in my own repository. I don't know why there were something wrong here.

0xpr03 (approved):
cc @dfaust I think we did an oopsi

0xpr03:
CI for BSD seems to fail because cirrus doesn't have enough capacity..

dfaust:
Yes, that's a mistake.

Thank you @CalunVier for fixing it!

mhnap:
Just in case, please when releasing update minor version, not patch. So it would be more visible that this is a breaking change. Thanks!

0xpr03:
Yeah, has to be for breaking change pre 1.0

PR #620, opened 2024-07-19
Suppress `dead_code` warnings on some platforms.

waywardmonkeys:
Not all platforms use the `BoundSender` and `bounded` and on those, building results in `dead_code` warnings.


dfaust:
Thanks for the PR. But I have fixed the issue by compiling `BoundSender` only on Windows and Linux, instead.

Issue #525, opened 2023-08-14
Breaking Changes v7


0xpr03:
List of things to change in v7

 - [ ] (maybe) PollWatcher: change timeout between intervals to include time spend performing work
 - [x] Config: Replace deprecated `poll_interval` signature with `poll_interval_v2`
 - [ ] (maybe) unify all places of panic/error for locking & thread starts
 - [ ] remove crossbeam
 - [ ] upgrade mio to 1.0 #560
 
 todo:
 
 - [ ] add nightly test with conditional compilation check for #592

zeroishero:
Is there more detailed information about 2 and 3?

0xpr03:
Not much. We introduced poll_intervall_v2 to stay semver compliant, we should change poll_intervall for v7 and thus remove v2.

And for thread start, join etc we don't handle errors the same way everywhere. Basic idea is to just decide whether to unwrap or error and do that everywhere.

zeroishero:
So, i checked the 2nd one. It seems poll_interval is returning Duration while poll_interval_v2 is returning Option<Duration>. So, the intended change is to just change poll_interval to return Option and change calls of poll_interval_v2 back to poll_interval? That seems easy. Or Am I missing something?

0xpr03:
Yep that's the plan for when we actually go for v7. Until then I'd leave it as is (in main), or we'll have a harder time making patch/fix releases.

zeroishero:
So, is there branch for v7? Will work related to this issue be accepted? I would like to do the 2nd one as it's quite easy and I will attempt 1.

0xpr03:
No we normally don't have that, but you can do a PR for these things anyway.

zeroishero:
Can you explain what the first checklist is about? I want to try that.

PR #622, opened 2024-07-19
Correctly specify `crossbeam-channel` as feature.

waywardmonkeys:
The feature controlling the usage of `crossbeam_channel` is not `crossbeam`.

This has been incorrect since it was made optional in be6ef8a1.

### DIFF
--- a/notify-debouncer-full/src/lib.rs
+++ b/notify-debouncer-full/src/lib.rs
@@ -123,7 +123,7 @@ where
     }
 }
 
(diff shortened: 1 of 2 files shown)


0xpr03:
Can be closed since crossbeam channel will go away.
#525

dfaust:
@0xpr03: I removed the internal use of crossbeam-channel, but it can still be used as a notification handler.

I also noticed this issue and fixed it in PR #630. I saw this PR only now.

Issue #581, opened 2024-03-27
PollWatcher: Ignore IO "file not found" exceptions when accessing entry metadata?🥺 


earshinov:
For local development, we are running our Django app inside a Linux Docker container, launched from Windows host with help of WSL2, with sources typically mounted from the Windows file system. We are using [watchfiles](https://github.com/samuelcolvin/watchfiles/) + [django_watchfiles](https://github.com/adamchainz/django-watchfiles) instead of Django's built-in reloader. Watchfiles uses `notify` under the hood ([source](https://github.com/samuelcolvin/watchfiles/blob/main/src/lib.rs)). Thanks to [this code in watchfiles](https://github.com/samuelcolvin/watchfiles/blob/9f19a18da5323d0e63a2ee16c47ed6c819f89be5/watchfiles/main.py#L337), it understands that it is launched from a Windows host and falls back to `PollWatcher` (👌).

We are facing an issue that `notify` hits an exception (also bringing down the Django app and the Docker container) when we launch `mypy` in the same running container:

```
_rust_notify.WatchfilesRustInternalError: error in underlying watcher: IO error for operation on /opt/project/.mypy_cache/3.10/iniconfig/__init__.meta.json.894e30fd1ecd2163: No such file or directory (os error 2) about ["/opt/project/.mypy_cache/3.10/iniconfig/__init__.meta.json.894e30fd1ecd2163"]
```

I am mediocre at best in Rust, but I assume that this exception is coming from here:
https://github.com/notify-rs/notify/blob/2511ebc004c3680bebc0a81f7eb8a7928ecdc76d/notify/src/poll.rs#L301

Could we perhaps have "Not found" errors ignored?

##### Software

watchfiles 0.21 (latest release)
notify 5.1.0 ([source](https://github.com/samuelcolvin/watchfiles/blob/v0.21.0/Cargo.lock#L172))

I see the same code in latest `notify` here in the repo, so I guess we would have the same problem with `watchfiles` rebuilt with latest `notify`.

##### OS

Host: Windows 11 x64 Version 23H2 (0S Build 22631.3296)

```
> wsl cat /proc/version
Linux version 5.10.102.1-microsoft-standard-WSL2 (oe-user@oe-host) (x86_64-msft-linux-gcc (GCC) 9.3.0, GNU ld (GNU Binutils) 2.34.0.20200220) #1 SMP Wed Mar 2 00:30:59 UTC 2022
```

Docker container: `python:3.11-bullseye-slim`

[...]

PR #634, opened 2024-08-26
Preserve IO errors in `PollWatcher`

samuelcolvin:
Part of solving #581.

In order to successfully process `io::Error`s in [watchfiles](https://github.com/samuelcolvin/watchfiles/), we need to preserve them as `ErrorKind::IO`, rather than using `ErrorKind::Generic`.

### DIFF
--- a/notify-types/src/event.rs
+++ b/notify-types/src/event.rs
@@ -195,7 +195,10 @@ pub enum RemoveKind {
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "kebab-case"))]
-#[cfg_attr(all(feature = "serde", not(feature = "serialization-compat-6")), serde(tag = "type"))]
+#[cfg_attr(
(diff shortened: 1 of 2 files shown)


notify-types/src/event.rs
```diff
@@ -195,7 +195,10 @@ pub enum RemoveKind {
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "kebab-case"))]
-#[cfg_attr(all(feature = "serde", not(feature = "serialization-compat-6")), serde(tag = "type"))]
+#[cfg_attr(
```
samuelcolvin, line 198: this is just running `cargo fmt`.

notify/src/poll.rs
```diff
@@ -259,25 +259,19 @@ mod data {
                 .follow_links(true)
                 .max_depth(Self::dir_scan_depth(is_recursive))
                 .into_iter()
-                //
-                // QUESTION: should we ignore IO Error?
```
samuelcolvin, line 263: This comment was left behind when #507 started returning IO Errors.

PR #640, opened 2024-09-17
Update all deps

allan2:
All tests with all features are passing.

clippy and fmt run.


dfaust:
Hey @allan2, thanks for the effort, but as you noticed, I will remove mock_instance, so it's easiest to close this PR.

PR #660, opened 2024-12-17
Fix `Event::kind` serialization with serialization-compat-6

Legend-Master:
The `kind` field in `Event` was serialized as `type` in v6, but is `kind` with `serialization-compat-6` feature

Reference: https://github.com/tauri-apps/plugins-workspace/issues/2221

I'm not sure how the release process work here, let me know if I need to add a change log entry (if so please also tell me what version should this change be in, since I see there's already a `notify-types 2.0.0 (unreleased)` entry) or do I need to target another branch

### DIFF
--- a/notify-types/src/event.rs
+++ b/notify-types/src/event.rs
@@ -307,6 +307,10 @@ pub struct Event {
         all(feature = "serde", not(feature = "serialization-compat-6")),
         serde(flatten)
     )]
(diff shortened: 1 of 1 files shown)


dfaust:
Thank you for the fix.
I'll merge it into `main` and create a new branch based on the last tag and then cherry-pick this commit.

PR #662, opened 2025-01-03
In notify-types, relicense src/event.rs to (MIT OR Apache-2.0)

musicinmybrain:
This file was copied from the main `notify` crate and had a header comment that indicated it was licensed `CC0-1.0` or `Artistic-2.0` with an additional clause (see https://github.com/notify-rs/notify/issues/514);

Since @dfaust indicates in https://github.com/notify-rs/notify/pull/661#discussion_r1901262603 that this file can be relicensed, we simply remove the old header comment to indicate that (by default) it falls under the overall `(MIT OR Apache-2.0)` license of the `notify-types` crate.

The [suggestion](https://github.com/notify-rs/notify/pull/661#discussion_r1901306684) to add `CC0-1.0` as an additional license option for `notify-types` could be done, if desired, as a follow-up PR.

### DIFF
--- a/notify-types/src/event.rs
+++ b/notify-types/src/event.rs
@@ -1,5 +1,3 @@
-// This file is dual-licensed under the Artistic License 2.0 as per the
-// LICENSE.ARTISTIC file, and the Creative Commons Zero 1.0 license.
 //! The `Event` type and the hierarchical `EventKind` descriptor.
 
 use std::{


dfaust:
Thanks.

musicinmybrain:
> The [suggestion](https://github.com/notify-rs/notify/pull/661#discussion_r1901306684) to add `CC0-1.0` as an additional license option for `notify-types` could be done, if desired, as a follow-up PR.

https://github.com/notify-rs/notify/pull/663

PR #663, opened 2025-01-03
Add a CC0-1.0 license option for the notify-types crate

musicinmybrain:
This changes the license of the `notify-types` crate from

```
MIT OR Apache-2.0
```

to

```
MIT OR Apache-2.0 OR CC0-1.0
```

and adds a copy of `LICENSE-CC0` from the main notify crate, as requested in https://github.com/notify-rs/notify/pull/661#discussion_r1901262603.

----

I don’t have a personal interest in this change – in fact, it makes my life *very slightly* more difficult as a downstream package maintainer, since `CC0-1.0` is not-allowed for code in Fedora and I would need to patch it out of the choices in the license expression. Still, this PR ties up the last loose end from the discussion in https://github.com/notify-rs/notify/pull/661.


dfaust:
> in fact, it makes my life very slightly more difficult as a downstream package maintainer, since CC0-1.0 is not-allowed for code in Fedora

In that case, let's not do it.
But what about `notify`? That crate is still licensed under CC0-1.0.

musicinmybrain:
> In that case, let's not do it. But what about `notify`? That crate is still licensed under CC0-1.0.

We have the following policy:

> Existing uses of CC0-1.0 on code files in Fedora packages prior to 2022-08-01, and subsequent upstream versions of those files in those packages, continue to be allowed. We encourage Fedora package maintainers to ask upstreams to relicense such files.

This exception covers our [`rust-notify` package](https://src.fedoraproject.org/rpms/rust-notify), which we have had since 2019-03-13, and I believe it covers `notify-types/src/event.rs` prior to https://github.com/notify-rs/notify/pull/662 since it was not a new `CC0-1.0`-licensed source file, but was simply moved or split out from the `notify` crate.

We would not be able to package the `notify` crate at all if we hadn’t done so prior to 2022-08-01.

*If* a relicensing of the remaining `CC0-1.0` code is possible (presumably with the consent of all relevant contributors), then that would certainly be welcome.

0xpr03:
There is an open issue, I fear notify is stuck in re licensing limbo hell.

PR #664, opened 2025-01-07
FileIdCache: Allow flexible handle instead of direct borrow for file ids

florian-g2:
Hi, 
I suggest changing the return type of the `cached_file_id` method in the `FileIdCache` from `Option<&FileId>` to `Option<impl Deref<Target=FileId>>;` (breaking change)

This allows trait consumers to choose more freely how to return the FileId. 
In a FileIdCache implementation I had to do I was required to return a owned copy of the file id. 
The changed return type allows eg. a Cow::Owned or any other new-type implementation that dereferences to a FileId.

This is now doable with the recent MSRV raise to 1.77.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## debouncer-full 0.6.0 (unreleased)
+- FEATURE: allow `FileIdCache` trait implementations to choose ownership of the returned file-ids
+
+## file-id 0.2.3 (unreleased)
+- CHANGE: implement `AsRef<FileId>` for `FileId`
+
 ## notify 8.0.0 (2025-01-10)
 
 - CHANGE: update notify-types to version 2.0.0
--- a/file-id/src/lib.rs
+++ b/file-id/src/lib.rs
@@ -106,6 +106,12 @@ impl FileId {
     }
 }
 
+impl AsRef<FileId> for FileId {
+    fn as_ref(&self) -> &FileId {
+        self
+    }
+}
+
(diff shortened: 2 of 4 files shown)


dfaust:
I'm split on this. `FileId` is just 256 bytes and implements `Copy`. So maybe returning `Option<FileId>` would be easiest. But I'm willing to merge it, if you think it's an improvement.
Please update the change-log as well.

florian-g2:
I changed the impl from `Deref<Target=FileId>` to `AsRef<FileId>`. The `AsRef` trait is more correct here.

I see no downside to allow a impl return. They are zero cost at runtime.
Using `impl AsRef<FileId>` allows the cache to return a owned value, a reference, a wrapper type or even a mutex guard.
In my codebase the FileId is wrapped in a new type to add some functionality, that was the motivation for this change.

As far as I see the change is backwards compatible. 
The trait will still allow implementations that return `Option<&FileId>` for the file id. 
Clippy will just show a `refining_impl_trait` warning.

florian-g2:
> Please update the change-log as well.

I am unfamiliar with the process. Should I just add the change at the top of the .md without a version / date?

dfaust:
> I am unfamiliar with the process. Should I just add the change at the top of the .md without a version / date?

Yes. Add it under `debouncer-full 0.6.0 (unreleased)`.

florian-g2:
Done. I opted for a simpler description for the change in the changelog. Note that `file-id` is also touched.

dfaust:
Thanks.

PR #669, opened 2025-01-25
remove unused imports from documentation

20jasper:
Removed unused imports and uncommented `Path` import for easy copy pasting


dfaust:
Thanks!

PR #673, opened 2025-02-19
feat: support wasm build

underfin:
We are using `notify` as watcher and build it to [wasm32-wasip2](https://doc.rust-lang.org/rustc/platform-support/wasm32-wasip2.html) and run it at https://stackblitz.com/. 

But when using `notify-debouncer-full` to instead of `notify`, it can't build because the `get_file_id` is not implement at wasm. The `std::os::wasi::fs::MetadataExt` is implement at nightly rust, the stable version hasn't it. So here using `NoCache` for wasm to avoid the issue.

I try to add a build wasm32-wasip2 ci, but the rust toolchain version is 1.77.2, it hasn't support the wasm32-wasip2 target.It need to bump rust toolchain at first. If you approved it, please ping me i will upgrade it at first.

Thank your excellent work!

### DIFF
--- a/notify-debouncer-full/src/cache.rs
+++ b/notify-debouncer-full/src/cache.rs
@@ -1,10 +1,6 @@
-use std::{
-    collections::HashMap,
-    path::{Path, PathBuf},
-};
-use file_id::{get_file_id, FileId};
+use file_id::FileId;
 use notify::RecursiveMode;
-use walkdir::WalkDir;
+use std::path::{Path, PathBuf};
 
 /// The interface of a file ID cache.
 ///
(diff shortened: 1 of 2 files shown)


notify-debouncer-full/src/lib.rs
```diff
@@ -79,7 +79,7 @@ use std::{
 
 use time::now;
 
-pub use cache::{FileIdCache, FileIdMap, NoCache, RecommendedCache};
+pub use cache::{FileIdCache, NoCache, RecommendedCache};
```
underfin, line 82: Here has a break change, but it should be recommend to user to using `RecommendedCache`.

notify-debouncer-full/src/cache.rs
```diff
@@ -38,57 +34,68 @@ pub trait FileIdCache {
...
-#[derive(Debug, Clone, Default)]
-pub struct FileIdMap {
-    paths: HashMap<PathBuf, FileId>,
-}
-
-impl FileIdMap {
-    /// Construct an empty cache.
-    pub fn new() -> Self {
-        Default::default()
+#[cfg(not(any(target_os = "linux", target_os = "android", target_family = "wasm")))]
```
dfaust, line 37:
```suggestion
#[cfg(not(target_family = "wasm"))]
```

This way we can still export the FileIdCache - except when we have a wasm target.

notify-debouncer-full/src/lib.rs
dfaust:
```suggestion
pub use cache::{FileIdCache, NoCache, RecommendedCache};

#[cfg(not(target_family = "wasm"))]
pub use file_id_map::FileIdMap;
```

notify-debouncer-full/src/cache.rs
```diff
@@ -38,57 +34,68 @@ pub trait FileIdCache {
...
+mod file_id_map {
```
dfaust, line 38: I would prefer this to be its own file. No need to make it a sub-module of `cache`, `crate::file_id_map` should be fine.

dfaust (changes requested):

dfaust:
> I try to add a build wasm32-wasip2 ci, but the rust toolchain version is 1.77.2, it hasn't support the wasm32-wasip2 target.It need to bump rust toolchain at first. If you approved it, please ping me i will upgrade it at first.

The CI setup for notify is totally broken at the moment. But ideally, it would run for multiple toolchains, regardless of the contents of the rust-toolchain.toml file. We want to run the pipelines for the nightly toolchain as well. If you want, you can add a wasm job.

I have never worked with wasi and I have no idea how notify should handle a wasi environment. How FS event notification even works there. Would it be possible to implement `get_file_id` for wasi?

underfin:
> How FS event notification even works there

The wasi implemented the basic std::fs relate API, it worked fine at https://stackblitz.com/.  
- ref https://github.com/rust-lang/rust/pull/59619
- ref https://doc.rust-lang.org/std/os/wasi/fs/index.html. 

> Would it be possible to implement get_file_id for wasi?

Yeah. The std::os::wasi::fs::MetadataExt is implement at nightly rust, the stable version hasn't it. If it move into stable version we could implement it.

PR #674, opened 2025-02-28
windows: Fix server hangs under some circumstance

JunkuiZhang:
Hi, thank you so much for creating such a high-quality Rust library.

We encountered a tricky issue while using this library in the development of Zed (https://github.com/zed-industries/zed). After testing, we identified the issue as a bug in `notify`. This PR provides a test case, `test_windows_trash_dir()`, which reproduces the bug.

The bug can be reproduced as follows: in a project directory `/path/to/project`, there is a subfolder `/path/to/project/child`. If we monitor changes in the child folder, then move the child folder to the trash (deleting the folder directly does not trigger the bug), and monitor the project directory, the server gets stuck at the `WaitForSingleObjectEx` command. 

This issue seems to be related to the `bAlertable` flag in `WaitForSingleObjectEx`, but we are still unclear why setting `bAlertable` to `TRUE` causes this bug.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,5 +45,6 @@ rstest = "0.24.0"
 serde = { version = "1.0.89", features = ["derive"] }
 serde_json = "1.0.39"
 tempfile = "3.10.0"
+trash = "5.2.2"
 walkdir = "2.4.0"
 windows-sys = "0.59.0"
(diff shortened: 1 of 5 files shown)


dfaust:
Hi @JunkuiZhang. Thanks so much for fixing this issue. The changes look good.
Unfortunately I don't have Windows to test. But I think we can merge this once the pipeline has succeeded. The pipeline works on the current main branch. So maybe a rebase will fix the issue.
Also, would you mind adding a change-log entry?

PR #676, opened 2025-03-02
Add `nighlty` feature for wasip2 support

dfaust:


xtqqczze:
Since `tempfile` is only a dev dependency, it should be fine to bump it to 3.23, where the `nightly` feature is no longer required for `wasip2` support. See the [`tempfile` changelog](https://github.com/Stebalien/tempfile/blob/master/CHANGELOG.md#3230).

PR #680, opened 2025-03-26
Add support for flume

RenDiego:
added basic support for [flume](https://crates.io/crates/flume)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
 # Changelog
 
+## notify 9.0.0 (unreleased)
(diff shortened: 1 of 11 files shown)


dfaust:
Thank you. The code looks good. The only thing that's missing is the documentation.
Basically, everywhere that `crossbeam-channel` is mentioned, it should also mention `flume`.
Plus, a change-log entry would be nice.

RenDiego:
I believe that's all,
I assumed the next notify version will be 9.0.0 instead of 8.0.1 and updated the changelog for debouncer-mini and debouncer-full

dfaust:
Thanks again!

PR #683, opened 2025-05-06
Do not remove a filename twice

### LINKED ISSUE #665: `kqueue` always returns an "No such file or directory" on `Watcher::unwatch`
#### System details

- OS/Platform name and version: macOS 13.0 Ventura
- Rust version (if building from source): `rustc --version`: rustc 1.81.0 (eeb90cda1 2024-09-04)
- Notify version (or commit hash if building from git): 7.0.0 with `macos_kqueue` feature flag
- If you're running as a privileged user (root, System): -
- If you're running in a container, details on the runtime and overlay: -
- If you're running in a VM, details on the hypervisor: -

I'm trying to use `kqueue` on macOS, but `Watcher::unwatch` always returns an error. Here is a minimal reproduction:

```
mkdir kqueue_repro && cd kqueue_repro
cargo init .
cargo add notify --features macos_kqueue
```

Source of the program:

``` rust
use std::error::Error;
use std::path::PathBuf;
use notify::{recommended_watcher, RecursiveMode, Watcher};

fn main() -> Result<(), Box<dyn Error>> {
    let path = PathBuf::from("src");

    let mut watcher = recommended_watcher(|event| println!("{:?}", event))?;
    watcher.watch(&path, RecursiveMode::Recursive)?;
    watcher.unwatch(&path)?; // <- This always returns "No such file or directory"

    Ok(())
}
```

Stepping through the source of `kqueue.rs`, and subsequently the `kqueue` crate, the error seems to be triggered in `Watcher::remove_filename`, specifically when calling `Watcher::delete_kevents`. What's also very weird is that when I try to set a break point in this function CodeLLDB goes absolutely nuts and allocates more than 40 GB before my computer goes poof.

#### What you expected

`Watcher::unwatch` does not throw.

#### What happened

```
Error: Error { kind: Generic("No such file or directory (os error 2) about [\".../src\"]"), paths: [] }
```

worr:
When unwatching a directory that was watched recursively, notify's kqueue impl would first unwatch the containing directory, then walk it to unwatch all the items inside.

However, walking the directory first processes the containing directory, and unwatching a directory twice can cause an error in the kqueue crate.

Fixes #665 and adds a test to prevent regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## notify 9.0.0 (unreleased)
 - FEATURE: added support for the [`flume`](https://docs.rs/flume) crate
+- FIX: kqueue-backend: do not double unwatch top-level directory when recursively unwatching [#683]
 
 ## debouncer-full 0.6.0 (unreleased)
 - FEATURE: allow `FileIdCache` trait implementations to choose ownership of the returned file-ids
--- a/notify/src/kqueue.rs
+++ b/notify/src/kqueue.rs
@@ -342,10 +342,6 @@ impl EventLoop {
         match self.watches.remove(&path) {
             None => return Err(Error::watch_not_found()),
             Some(is_recursive) => {
-                self.kqueue
-                    .remove_filename(&path, EventFilter::EVFILT_VNODE)
-                    .map_err(|e| Error::io(e).add_path(path.clone()))?;
-
                 if is_recursive || remove_recursive {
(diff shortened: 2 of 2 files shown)


dfaust:
Thanks a bunch. Just one thing. Would you mind adding it to the change log?

worr:
@dfaust Done, let me know if there are issues with the format or anything you want me to change

dfaust:
Thank you!

PR #684, opened 2025-05-06
Bump kqueue to 1.1.1

worr:
This release fixes some file descriptor leaks, reduces memory usage and also fixes a bug where unknown filenames would try and unwatch fd 0.

---

btw, feel free to tag me on kqueue-backend bugs, since I maintain the upstream kqueue lib


dfaust:
Thanks.

> btw, feel free to tag me on kqueue-backend bugs, since I maintain the upstream kqueue lib. I can't promise I'll always be super quick to respond, but I'd be happy to take a look and help if I'm able.

I'll try to remember.

PR #692, opened 2025-06-26
feat: introduce `Watcher::paths_mut` for adding/removing paths in batch

### LINKED ISSUE #653: [Feature Request]: provider a watch multiple files api to instead of multiple call `Watcher#watch`
#### System details

- OS/Platform name and version: MacOS 14.4.1 (M3 pro)
- Rust version (if building from source): `rustc --version`: rustc 1.77.2
- Notify version (or commit hash if building from git): 7.0.0

- If you're coming from a project that makes use of Notify, what it is, and a link to the downstream issue if there is one:
- Filesystem type and options:
- On Linux: Kernel version:
- On Windows: version and if you're running under Windows, Cygwin (unsupported), Linux Subsystem:
- If you're running as a privileged user (root, System):
- If you're running in a container, details on the runtime and overlay:
- If you're running in a VM, details on the hypervisor:

The performance has more overhead if multiple call `Watcher#watch`, watching 158 files take 88 ms at my local.  The overhead look like the `FsEventWatcher#stop` and `FsEventWatcher#run` is call multiple times. It could be improve like this. 
``` .rs

fn watch_multiple(&mut self, path: Vec<&Path>, recursive_mode: RecursiveMode) -> Result<()> {
        self.stop();
        for path in path {
                   let result = self.append_path(path, recursive_mode);
       }
        // ignore return error: may be empty path list
        let _ = self.run();
        result
    }
```

branchseer:
fixes #653.

I can add tests once the added API is agreed upon.

### DIFF
--- a/notify/src/lib.rs
+++ b/notify/src/lib.rs
@@ -566,21 +566,21 @@ mod tests {
         fs::write(&b_file1, b"Lorem ipsum")?;
 
         // wait for create events of a/file1 and b/file1
-        let mut a_file1_encoutered: bool = false;
-        let mut b_file1_encoutered: bool = false;
+        let mut a_file1_encountered: bool = false;
+        let mut b_file1_encountered: bool = false;
         for event in iter_with_timeout(&rx) {
             for path in event.paths {
-                a_file1_encoutered =
-                    a_file1_encoutered || (path == a_file1 || path == a_file1.canonicalize()?);
-                b_file1_encoutered =
-                    b_file1_encoutered || (path == b_file1 || path == b_file1.canonicalize()?);
+                a_file1_encountered =
+                    a_file1_encountered || (path == a_file1 || path == a_file1.canonicalize()?);
+                b_file1_encountered =
+                    b_file1_encountered || (path == b_file1 || path == b_file1.canonicalize()?);
             }
-            if a_file1_encoutered && b_file1_encoutered {
+            if a_file1_encountered && b_file1_encountered {
                 break;
             }
         }
-        assert!(a_file1_encoutered, "Did not receive event of {a_file1:?}");
-        assert!(b_file1_encoutered, "Did not receive event of {b_file1:?}");
+        assert!(a_file1_encountered, "Did not receive event of {a_file1:?}");
+        assert!(b_file1_encountered, "Did not receive event of {b_file1:?}");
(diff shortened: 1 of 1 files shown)


BoomchainLabs (approved):

branchseer:
@JohnTitor Test & example added

JohnTitor:
Seems there are some typos, check CI.

JohnTitor (approved):
Great, thank you!

TimNN:
Thanks! But I wanted to double-check that this handles all edge cases appropriately:

* What happens if `commit` is called twice?
* What happens if `add` or `remove` are called after `commit`?
* What happens if `PathsMut` is dropped without calling commit?

Boshen:
Thank you for merging this.

@dfaust will you be able to publish a new release?

TimNN:
Ack, thanks for clarifying @branchseer!

For FSEvents, I was mainly concerned that calling `fn run` while the watcher is already running could cause problems, or that modifying the watched paths while the watcher is running could be problematic.

For example, `fn stop` includes some code to explicitly terminate the previous `runloop`.

If `fn run` is called twice (e.g. due to `fn commit` being called twice), it looks like that termination code doesn't run. (Maybe that's fine, I haven't tested it and am not familiar enough with the code to properly judge this).

Also: `paths` seems to be passed to `FSEventStreamCreate` but also modified in `fn append_path`. Calling `fn add` after `fn commit` means that's it's now possible for `paths` to be modified while the watcher is running. (Again, maybe that's fine (I couldn't tell from the Apple documentation) because `FSEventStreamCreate` makes a copy of the paths or only reads them during initialization, but I could also easily see this cause problems).

branchseer:
There's a `stop` call in `FsEventPathsMut::new`: https://github.com/notify-rs/notify/blob/d824023e483b24ceb398f08dec20f35ce575e7fa/notify/src/fsevent.rs#L273

and since `FsEventPathsMut` mutably borrows the watcher, we can safely assume the watcher is not started when a `FsEventPathsMut` exists.

branchseer:
I didn't notice that `run` and `stop` are unbalanced. Thanks for pointing it out. @TimNN 

I have submitted a fix: https://github.com/notify-rs/notify/pull/695/

PR #693, opened 2025-06-30
Do not clone paths while walking dirs

riberk:
To avoid unneeded allocations, this commit uses `walkdir::DirEntry::into_path` instead


JohnTitor (approved):
Thanks!

PR #695, opened 2025-07-02
fix: make `PathsMut::commit` consuming

branchseer:
Ensures `PathsMut::commit` can only be called once, and adds more docs.
fixes concerns at https://github.com/notify-rs/notify/pull/692#issuecomment-3026502572


TimNN:
Thanks @branchseer, this addresses all the concerns I had with the previous implementation!

PR #698, opened 2025-07-15
feat(inotify): notify a user if the `max_user_watches` has been reached implicitly

riberk:
There was no way to detect the problem if it caused by implicitly watched directory (e.g. recursive watch). This commit gives a user the ability to handle the problem and react somehow.

For example, gui app can show a notification with some instructions how to increase the value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## notify 8.2.0 (unreleased)
+- FEATURE: notify a user if the `max_user_watches` has been reached implicitly (`INotifyWatcher`) [#698]
+
+[#698]: https://github.com/notify-rs/notify/pull/698
+
 ## notify 8.1.0 (2025-07-03)
 - FEATURE: added support for the [`flume`](https://docs.rs/flume) crate
 - FIX: kqueue-backend: do not double unwatch top-level directory when recursively unwatching [#683]


riberk:
> I just wish we could run it in the CI because I can see it easily being forgotten.

I thought about it, but it requires root, and I just don't know how to solve the problem, because I did not delve into how CI works in the project.
* If CI runs with root privileges, then I can just add runtime check if the user is root or not and run the test only for the root user.
* If CI runs WITHOUT the privileges, then it requires more complex solution with some kind of containerization / virtualization, and I think it doesn't worth it

> Can you please add a change-log entry.
Yep, no problem

riberk:
> Yep, no problem

Ok, there is a problem:)

Which version of `notify` should it be?

There is already released `8.1.0` in [CHANGELOG.md](https://github.com/notify-rs/notify/blob/main/CHANGELOG.md#changelog)

`notify 8.2.0 (unreleased)`?

dfaust:
Yes, `notify 8.2.0` is good. Or leave it empty.

dfaust:
Thank you!

> I thought about it, but it requires root, and I just don't know how to solve the problem, because I did not delve into how CI works in the project.
> 
>     * If CI runs with root privileges, then I can just add runtime check if the user is root or not and run the test only for the root user.
> 
>     * If CI runs WITHOUT the privileges, then it requires more complex solution with some kind of containerization / virtualization, and I think it doesn't worth it

I don't know much about GitHub Actions, but after a quick search it looks like there is no root access and containerization is the only way.

PR #719, opened 2025-10-19
raise MSRV to 1.82

riberk:
This PR updates MSRV to 1.82 (min allowed) because of the [indexmap crate update](https://github.com/indexmap-rs/indexmap/pull/422). This crate is a dependency of a `proc-macro-crate` which is a dependency of `rs-test`.

MSRV before the change is breaking like in [this](https://github.com/notify-rs/notify/actions/runs/18627493522/job/53107493756) CI failure


riberk:
@JohnTitor Thanks for the review!

> You should also update the version string in Cargo.toml:

Done, and have updated README.md either

PR #720, opened 2025-10-20
Fix: INotifyWatcher keeps watching deleted paths

### LINKED ISSUE #709: Inotify: Removing a subdirectory in a recursively watched directory and then unwatching causes an error
#### System details

- OS/Platform name and version: Arch Linux
- Rust version (if building from source): `rustc --version`: rustc 1.89.0 (29483883e 2025-08-04)
- Notify version (or commit hash if building from git): 8.2.0
- On Linux: Kernel version: 6.16.1-arch1-1

1. Recursively watch directory `a`
2. Remove subdirectory `b`
3. Unwatch directory `a`

Code to reproduce:
```rust
use notify::{Config, INotifyWatcher, RecursiveMode, Result, Watcher};
use std::path::PathBuf;

fn main() -> Result<()> {
    let path = PathBuf::from(".");
    let subdirectory_path = path.join("some_dir");
    std::fs::create_dir(&subdirectory_path)?;

    let mut watcher = INotifyWatcher::new(|_| (), Config::default())?;
    watcher.watch(&path, RecursiveMode::Recursive)?;
    std::fs::remove_dir(subdirectory_path)?;
    watcher.unwatch(&path).unwrap();

    Ok(())
}
```

#### What you expected
No crash

#### What happened
Crashed with the output:
```
thread 'main' panicked at src/main.rs:12:28:
called `Result::unwrap()` on an `Err` value: Error { kind: Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" }), paths: ["/home/username/test/./some_dir"] }
```

When the subdirectory gets removed, there is an [attempt to remove its descriptor from the inotify instance](https://github.com/notify-rs/notify/blob/9c0ae15c9431c48c3de466f9c9aa9cb126cc48bc/notify/src/inotify.rs#L481-L484) but this fails since it has already been automatically removed and no longer exists. Consequently, the function returns early and doesn't execute `self.paths.remove(&w);`, leaving the subdirectory descriptor in `self.paths`. This error is [discarded](https://github.com/notify-rs/notify/blob/9c0ae15c9431c48c3de466f9c9aa9cb126cc48bc/notify/src/inotify.rs#L380).

When unwatching the parent directory, there is [another attempt to remove it](https://github.com/notify-rs/notify/blob/9c0ae15c9431c48c3de466f9c9aa9cb126cc48bc/notify/src/inotify.rs#L489-L492). This time the error is not discarded and the app panics.

riberk:
This PR fixes #709. Full information about the bug are described in the documentation for the `remove_single_descriptor` method.

The main problem was, that we tried to remove descriptor from the `inotify.watches()`, and, if it was completed successfully, it would be deleted from the hash `descriptor -> path`. But after FS entry deletion its file descriptor is invalidated and causes an error, which interrupts deletion from the hash

### DIFF
--- a/notify/src/inotify.rs
+++ b/notify/src/inotify.rs
@@ -818,6 +818,8 @@ mod tests {
             matches!(unwatch_result, Ok(())),
             "error: {unwatch_result:#?}"
         );
+    }
+
     #[test]
     fn create_file() {
         let tmpdir = testdir();


JohnTitor (approved):
Thank you and sorry for the delay! LGTM.

PR #722, opened 2025-10-21
Fix: Enable Windows to determine File or Folder in events

### LINKED ISSUE #261: CreateKind | RemoveKind | ModifyKind seem to always be `Any` most of the time?
#### System details

- OS/Platform name and version: Windows 10
- Rust version (if building from source): `1.46.0`
- Notify version (or commit hash if building from git): `5.0.0-pre.3`

- Filesystem type and options: NTFS

I have a basic file watcher function which watch for any event from some `path` recursively:
```rust
let mut source_watcher = immediate_watcher(move |res| match res {
	Ok(event) => {
		let event: Event = event;
		log::info!("{:#?}", event);
	}
	Err(e) => log::error!("{}", e),
})?;

source_watcher.watch(&path, RecursiveMode::Recursive)?;
// ...
```

#### What you expected

The log message will print a helpful message of what exactly happened with the file/folder under that path.

#### What happened

The log message will almost always print an `Any` variant of `CreateKind`, `RemoveKind`, and `ModifyKind`. Only when I try renaming some file that it manage to print the `ModifyKind::Name` variant.

Wouldn't this be quite useless for a more complex situation where you need to know what exactly happened to the file/folder?

boenrobot:
Replace ReadDirectoryChangesW with ReadDirectoryChangesExW, out of which the FileAttributes can tell if a file or folder is being created/modified/deleted.

This raises the minimum required Windows version to Windows 10, as this is where ReadDirectoryChangesExW was introduced.

Fixes #261


notify/src/windows.rs
```diff
@@ -422,44 +424,49 @@ unsafe extern "system" fn handle_event(
...
-                    FILE_ACTION_MODIFIED => {
-                        let kind = EventKind::Modify(ModifyKind::Any);
-                        let ev = newe.set_kind(kind);
-                        event_handler(Ok(ev));
-                    }
-                    _ => (),
-                };
-            }
+            match cur_entry.Action {
+                FILE_ACTION_RENAMED_OLD_NAME => {
```
boenrobot: I also rolled the FILE_ACTION_RENAMED_OLD_NAME into the match... Because I can't see any reason it shouldn't be in it already. Was there a good reason that I'm unaware of? If so, I can revert this part.

notify/src/windows.rs
```diff
@@ -303,28 +331,198 @@ fn start_read(
...
+
+                if ret == 0 {
+                    // error reading. retransmute request memory to allow drop.
+                    // Because of the error, ownership of the `overlapped` alloc was not passed
+                    // over to `ReadDirectoryChangesExW`.
+                    // So we can claim ownership back.
+                    let _overlapped = Box::from_raw(overlapped);
+                    let request = Box::from_raw(request);
+                    ReleaseSemaphore(request.data.complete_sem, 1, ptr::null_mut());
+                }
```
JohnTitor: I think it could fail silently, right? If so, the watched paths are silently ignored and it may cause some confusion. Some form of logging would be nice.

PR #723, opened 2025-10-24
raise MSRV to 1.85

riberk:
Ok, after the last MSRV change 4 days ago, there has been [new error](https://github.com/notify-rs/notify/actions/runs/18785065449/job/53601117441?pr=705). The crate `trash` for a windows watcher [was updated](https://github.com/Byron/trash-rs/commit/1894bfe4ab6677ca833cc60bbd6e568480abbf77#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R10-R11) to use 1.85. Fortunately, we haven't released the major update yet

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,9 +14,9 @@ task:
     - name: FreeBSD 14.0 - Rust nightly
       env:
         RUST_VERSION: nightly
-    - name: FreeBSD 14.0 - Rust 1.82 (MSRV)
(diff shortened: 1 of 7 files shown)


JohnTitor (approved):
Maybe it's a good time to consider migrating from a fixed MSRV policy to a flexible one (last N versions), to avoid MSRV bump requires a major bump.

PR #726, opened 2025-10-28
Use `objc2-core-foundation` and `objc2-core-services`

madsmtm:
`objc2-core-foundation` provides safer and auto-generated wrappers around CoreFoundation than the ones provided by `fsevent-sys`. In particular, you no longer have to worry about when to call `CFRelease`, `objc2-core-foundation` will mostly take care of that for you.

`objc2-core-services` does the same, but for CoreServices, in particular the FSEvents API that `notify` uses.

I maintain both crates as part of [the `objc2` project](https://github.com/madsmtm/objc2).

I have tried to keep the functionality as close to the existing as possible, to make this PR easier to review and bisect, though I suspect the code in `append_path` and `remove_path` could be simplified.

This is an alternative to updating to `fsevent-sys v5.1.0`. I have filed a similar PR to `fsevent` in https://github.com/octplane/fsevent-rust/pull/48.

### DIFF
--- a/notify/src/fsevent.rs
+++ b/notify/src/fsevent.rs
@@ -460,7 +460,8 @@ impl FsEventWatcher {
                     // Safety:
                     // This may panic if OOM occurs.
                     // Related: https://github.com/madsmtm/objc2/issues/797
-                    let cur_runloop = cf::CFRunLoop::current().expect("Failed to get current runloop");
+                    let cur_runloop =
+                        cf::CFRunLoop::current().expect("Failed to get current runloop");
 
                     #[allow(deprecated)]
                     fs::FSEventStreamScheduleWithRunLoop(


riberk:
I'm not a member, but I noticed your suggestion to replace [servo core foundation](https://github.com/servo/core-foundation-rs/issues/729) with these crates (the team has accepted the changes, but they haven't been merged yet), and also the [wgpu team](https://github.com/gfx-rs/wgpu/pull/5641) did the same.

I think it’s worth mentioning, since they have a much larger and more active community, and they might have already investigated these crates.

IMO, it’s a good idea, but personally I’d prefer us to have more tests — especially for corner cases — **before** applying the update.

notify/src/fsevent.rs
```diff
@@ -466,21 +456,22 @@ impl FsEventWatcher {
                 let stream = stream.0;
 
                 unsafe {
-                    let cur_runloop = cf::CFRunLoopGetCurrent();
+                    let cur_runloop = cf::CFRunLoop::current().unwrap();
```
riberk: Is it safe because we are sure, that the current run loop exists?

riberk:
> Fair enough, how do you propose we test things like this though? It's kinda hard, since most of it touches the file system and involves multiple processes to test their interaction.

OK, I'm going to add tests from [the old PR](https://github.com/notify-rs/notify/pull/415) and others, if something else comes to my mind

riberk:
> @riberk You are very welcome to do so. But be careful, these tests used to be extremely flaky. That's why we removed them at some point. In order to be useful we need automatic retries.

Yeah, I got it. I think I’ll be able to adapt them so they won’t be flaky — I’ll use channels instead of timeouts, ignore the order of events, and so on.

madsmtm:
`cargo test` passes on macOS 15.6.1 if that's all you wanted me to test?

JohnTitor:
Yeah, thank you!
I think the tests added by @riberk would be a great safeguard here.

Basically I'd like to accept this replacement, given the community leans to accept transitions and the benefits from it.
I'm going to take a deeper look in a week or so.

PR #767, opened 2026-01-26
Implement EventHandler for futures and tokio UnboundedSender

riberk:
## Description
This PR implements `EventHandler` (notify, debouncer-full and mini) for tokio and futures UnboundedSender behind the feature flags `futures` and `tokio`

### DIFF
--- a/notify/src/poll.rs
+++ b/notify/src/poll.rs
@@ -692,10 +692,6 @@ impl Drop for PollWatcher {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashSet;
-
-    use walkdir::WalkDir;
-
     use super::PollWatcher;
     use crate::test::*;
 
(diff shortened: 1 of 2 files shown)


notify-debouncer-full/src/lib.rs
```diff
@@ -956,4 +970,48 @@ mod tests {
...
+        fs::write(&file_path, b"Lorem ipsum").unwrap();
+
+        tokio::time::timeout(Duration::from_secs(10), rx.next())
+            .await
+            .expect("timeout")
+            .expect("No event")
+            .expect("error");
+    }
+
+    #[cfg(feature = "futures")]
```
JohnTitor: Should be:

```suggestion
    #[cfg(feature = "tokio")]
```

Other tests for tokio has the same left-over.

riberk: Thanks! Fixed

Additionally, tried to fix other PollWatcher tests

PR #887, opened 2026-04-11
Add `must_use` annotation to config builder methods

lukesandberg:
## Description
Add `must_use` annodations to the `with_*` functions on `Config`.  This way rust will warn if callers fail to use the result

## Related Issues
https://github.com/vercel/next.js/pull/92631 fixed a long standing bug in next.js because we failed to capture this return value


JohnTitor (approved):
Makes sense, thanks!

PR #902, opened 2026-04-19
Emit `remove` events even if a file was created and then removed

### LINKED ISSUE #900: [Bug]: fsevents are not delivering remove events if there is create and modify in the queue
### Prerequisites

### Component

notify (core library)

### Operating System

macOS

### Notify Version

notify-debouncer-full 0.8.1

### Notify Backend

Unknown/Auto-detected

### Bug Report

Here are some tests to verify the behavior

```rs
    ///  if macOS, FSEvents reports file deletions as a burst
    /// of `Create(File)` + `Modify(Data)` + `Remove(File)`.  The debouncer
    /// must not suppress the `Remove` event, even though a prior `Create`
    /// for the same path exists in the queue.
    #[test]
    fn remove_event_not_swallowed_after_create() -> Result<(), Box<dyn std::error::Error>> {
        let dir = tempdir()?;
        let dir_path = dir.path().canonicalize()?;

        let (tx, rx) = std::sync::mpsc::channel();
        let mut debouncer = new_debouncer(Duration::from_millis(10), None, tx)?;
        debouncer.watch(&dir_path, RecursiveMode::NonRecursive)?;

        // Create a file and wait for the debouncer to deliver the Create event.
        let file_path = dir_path.join("ephemeral.txt");
        fs::write(&file_path, b"will be deleted")?;

        let deadline = Instant::now() + Duration::from_secs(5);
        let mut got_create = false;
        while Instant::now() < deadline {
            match rx.recv_timeout(Duration::from_millis(100)) {
                Ok(Ok(events)) => {
                    if events.iter().any(|e| {
                        matches!(e.event.kind, EventKind::Create(_))
                            && e.event.paths.contains(&file_path)
                    }) {
                        got_create = true;
                        break;
                    }
                }
                Ok(Err(_)) => {}
                Err(std::sync::mpsc::RecvTimeoutError::Timeout) => continue,
                Err(_) => break,
            }
        }
        assert!(got_create, "expected Create event for ephemeral.txt");

        // Drain any remaining events from the creation.
        std::thread::sleep(Duration::from_millis(200));
        while rx.try_recv().is_ok() {}

        // Delete the file.
        fs::remove_file(&file_path)?;

        // The debouncer MUST deliver an event for this path (Remove, or at
        // minimum any event whose path matches so the consumer can stat it).
        let deadline = Instant::now() + Duration::from_secs(5);
        let mut got_removal = false;
        while Instant::now() < deadline {
            match rx.recv_timeout(Duration::from_millis(100)) {
                Ok(Ok(events)) => {
                    if events.iter().any(|e| e.event.paths.contains(&file_path)) {
                        got_removal = true;
                        break;
                    }
                }
                Ok(Err(_)) => {}
                Err(std::sync::mpsc::RecvTimeoutError::Timeout) => continue,
                Err(_) => break,
            }
        }
        assert!(
            got_removal,
            "expected Remove (or any) event for deleted ephemeral.txt, got none within 5s"
        );

        Ok(())
    }

    /// Unit-level reproducer for the same bug: feed a Create + Remove
    /// sequence into `DebounceDataInner` directly, with the queue already
    /// flushed between them (simulating the debounce tick).  The Remove
    /// must not be swallowed.
    #[test]
    fn push_remove_after_flushed_create() {
        use crate::NoCache;
        use std::path::PathBuf;
        use notify::event::{CreateKind, RemoveKind};

        let mut state = DebounceDataInner::new(NoCache, Duration::from_millis(50));

        let time = std::time::Instant::now();
        MockTime::set_time(time);

        let path = PathBuf::from("/tmp/test_file.txt");

        // Simulate: file created → Create event added
        state.add_event(Event {
            kind: EventKind::Create(CreateKind::File),
            paths: vec![path.clone()],
            ..Default::default()
        });

        // Simulate: debounce tick flushes the Create
        MockTime::advance(Duration::from_millis(100));
        let flushed = state.debounced_events();
        assert_eq!(flushed.len(), 1);
        assert!(matches!(flushed[0].event.kind, EventKind::Create(_)));

        // Simulate: FSEvents sends Create + Remove for the deletion
        // (this is what macOS does)
        state.add_event(Event {
            kind: EventKind::Create(CreateKind::File),
            paths: vec![path.clone()],
            ..Default::default()
        });
        state.add_event(Event {
            kind: EventKind::Remove(RemoveKind::File),
            paths: vec![path.clone()],
            ..Default::default()
        });

        // Flush again — we MUST get the Remove event
        MockTime::advance(Duration::from_millis(100));
        let flushed = state.debounced_events();
        assert!(
            flushed
                .iter()
                .any(|e| matches!(e.event.kind, EventKind::Remove(_))),
            "expected Remove event after flushed Create, got: {flushed:?}"
        );
    }
```

The problem that I see is that remove events are swallewed by your queue dedup and not produced at all, my fix is breakign some of your tests but the tests makes very littel sense to me

```
    tests::state::file_name_03___add_create_dir_event_twice__
    tests::state::file_name_26___add_remove_event_after_create_event__
    tests::state::file_name_28___add_remove_event_after_create_and_modify_event__
    tests::watched_paths_with_watch_update_paths_and_unwatch
```

in short - if the remove event is produced I need to see it, inotify seems to be working fine but fsevents -no

dfaust:
Always emit `remove` events, regardless if files were just created.

## Description
Currently debouncer-full ignores `create` + `remove` events for a file. But @dmtrKovalenko reports that fsevents emits `create` + `remove` for remove events. So this PR changes the behavior to always emit `remove` events, regardless if the files were just created. (the `create` event is still not emitted in such a case)

I don't have macOS, so I can't test this. We could add a CI test, but I'm not sure if that is worth it.

@JohnTitor  This is a breaking change, so you may want to release debouncer-full 0.8.0 before merging.

## Related Issues
Fix https://github.com/notify-rs/notify/issues/900

### DIFF
--- a/notify-debouncer-full/CHANGELOG.md
+++ b/notify-debouncer-full/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
+
 ## unreleased
+
+- CHANGE: emit `remove` events event if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
+
+[#900]: https://github.com/notify-rs/notify/issues/900
 
 ## debouncer-full 0.8.0-rc.1 (2026-04-16)
 
--- a/notify-debouncer-full/CHANGELOG.md
+++ b/notify-debouncer-full/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-- CHANGE: emit `remove` events event if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
+- CHANGE: emit `remove` events even if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
 
 [#900]: https://github.com/notify-rs/notify/issues/900
 
--- a/notify-debouncer-full/src/lib.rs
+++ b/notify-debouncer-full/src/lib.rs
@@ -509,9 +509,6 @@ impl<T: FileIdCache> DebounceDataInner<T> {
         self.cache.remove_path(path);
 
         match self.queues.get_mut(path) {
-            Some(queue) if queue.was_created() => {
-                self.queues.remove(path);
-            }
             Some(queue) => {
                 queue.events = [DebouncedEvent::new(event, time)].into();
             }
@@ -963,6 +960,7 @@ mod tests {
     #[rstest]
     fn state(
         #[values(
+            "add_create_and_remove_event",
             "add_create_event",
             "add_create_event_after_remove_event",
             "add_create_dir_event_twice",
--- a/notify-debouncer-full/test_cases/add_create_and_remove_event.hjson
+++ b/notify-debouncer-full/test_cases/add_create_and_remove_event.hjson
@@ -8,12 +8,12 @@
             /watch/file: {
                 events: [
                     { kind: "create-any", paths: ["*"] }
-                    { kind: "modify-data-any", paths: ["*"] }
                 ]
             }
         }
     }
     events: [
+        { kind: "create-any", paths: ["/watch/file"] }
         { kind: "remove-any", paths: ["/watch/file"] }
     ]
     expected: {
--- a/notify-debouncer-full/test_cases/add_remove_event_after_create_event.hjson
+++ b/notify-debouncer-full/test_cases/add_remove_event_after_create_event.hjson
@@ -1,3 +1,7 @@
+// https://github.com/notify-rs/notify/issues/900
+//
+// macOS repeats the "create" event when a file is created and then removed.
+// The "remove" event must not be swallowed.
 {
     state: {
         queues: {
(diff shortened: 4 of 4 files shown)


notify-debouncer-full/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
+
 ## unreleased
 
+- CHANGE: emit `remove` events event if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
```
JohnTitor:
```suggestion
- CHANGE: emit `remove` events even if a file was created and then removed (because macOS repeats the "create" event) [#900] **breaking**
```

JohnTitor (approved):
Looks good!

JohnTitor:
> This is a breaking change, so you may want to release debouncer-full 0.8.0 before merging.

Let's bring this into the v0.8 window, notify v9 GA would still require some more time anyway and it's a good time to introduce breaking changes.

PR #905, opened 2026-04-24
perf: skip construction for unchanged stuff

JohnTitor:
## Description

Optimize `PollWatcher` rescans by deferring reported-path construction until an event is actually emitted.
The rescan loop now compares the existing and new `PathData` first, updates stored data in place for existing paths, and only builds the reported `Event` path when `compare_to_kind` returns an actual event kind.

On my end, it's roughly a 25% improvement for the measured no-op recursive polling workload.

### DIFF
--- a/notify/src/poll.rs
+++ b/notify/src/poll.rs
@@ -249,17 +249,23 @@ mod data {
                 self.follow_symlinks,
                 false,
             ) {
-                let old_path_data = self
-                    .all_path_data
-                    .insert(path.clone(), new_path_data.clone());
+                let event_kind = if let Some(old_path_data) = self.all_path_data.get_mut(&path) {
+                    let event_kind =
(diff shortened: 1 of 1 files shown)


notify/src/poll.rs
```diff
@@ -249,17 +249,23 @@ mod data {
...
-                );
-                if let Some(event) = event {
+                let event_kind = if let Some(old_path_data) = self.all_path_data.get_mut(&path) {
+                    let event_kind =
+                        PathData::compare_to_kind(Some(&*old_path_data), Some(&new_path_data));
+                    *old_path_data = new_path_data;
+                    event_kind
+                } else {
+                    let event_kind = PathData::compare_to_kind(None, Some(&new_path_data));
+                    self.all_path_data.insert(path.clone(), new_path_data);
```
JohnTitor, line 259: The main win comes from not calling this clone unconditionally.

PR #906, opened 2026-04-25
fix: replace watcher when rewatching

### LINKED ISSUE #708: Clarify what happens if you multi-watch a directory.
I'd like to know what happens in cases such as

```rs
let watcher: RecommendedWatcher = ...
watcher.watch(dir, RecursiveMode::Recursive);
watcher.watch(dir, RecursiveMode::NonRecursive);
```

Does it override the previous watching mode, or do nothing, or what?

Currently I can't see any information in the documentation about the behavior which makes me have to overly-aggressively unwatch things.

JohnTitor:
## Description

Fix https://github.com/notify-rs/notify/issues/708

Currently, rewatching behavior varies by backend and brings some confusion like #708.
This makes it consistent over all the backend and documents its behavior.

### DIFF
--- a/docs/UPGRADING_V8_TO_V9.md
+++ b/docs/UPGRADING_V8_TO_V9.md
@@ -76,6 +76,34 @@ let path = std::env::current_dir()?.join("src");
 watcher.watch(&path, notify::RecursiveMode::Recursive)?;
 ```
 
+### 4) Rewatching the same path replaces the existing watch
+
+Calling `Watcher::watch` again for the same backend-resolved path now replaces the existing watch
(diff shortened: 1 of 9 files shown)


dfaust:
Looks good. Although I would prefer if you could add some comments to the `inotify` and `kqueue` code, as well as `recursive_user_watch_ancestor`. Those parts are non-trivial and I did not verify them in detail. But the tests look good.

JohnTitor:
Makes sense, added in https://github.com/notify-rs/notify/pull/906/changes/f2279373a0645541499e60c40d6e66bc2f16ca54

PR #907, opened 2026-04-26
fix(kqueue): do not infer file paths on non-recursive watcher

### LINKED ISSUE #644: Non-recursive `KqueueWatcher` reports wrong event paths
#### System details

- OS/Platform name and version: `FreeBSD freebsd 14.1-RELEASE FreeBSD 14.1-RELEASE releng/14.1-n267679-10e31f0946d8 GENERIC amd64` (running under QEMU)
- Rust version (if building from source): `rustc --version`: `rustc 1.81.0 (eeb90cda1 2024-09-04)`
- Notify version (or commit hash if building from git): 6.1.1

Note: the issue doesn't seem to be FreeBSD-specific (as far as I can tell, it happens on all platforms that use kqueue).

- If you're coming from a project that makes use of Notify, what it is, and a link to the downstream issue if there is one: in [Arti](https://gitlab.torproject.org/tpo/core/arti), we're using `notify` to watch configuration files for changes.  The bug described here is causing [one of our tests to stall](https://gitlab.torproject.org/tpo/core/arti/-/blob/075101475ee929960788bb0b3bcbfbcb1d784126/crates/tor-config/src/file_watcher.rs#L529-557) to stall: https://gitlab.torproject.org/tpo/core/arti/-/issues/1644

I set up a non-recursive watcher for a directory that contains a single file `foo`. I created another file `bar` in the directory, expecting to receive an `Event` associated with the path of `bar`, but instead got an event for `foo`.

```rust
use notify::{Config, RecommendedWatcher, RecursiveMode, Watcher};
use std::path::PathBuf;
use std::sync::mpsc;
use temp_dir::TempDir;

fn write_file(dir: &TempDir, name: &str, data: &[u8]) -> PathBuf {
    let path = dir.path().join(name);
    std::fs::write(&path, data).unwrap();
    path
}

fn main() {
    let temp_dir = TempDir::new().unwrap();

    // Prepopulate temp_dir with a file.
    let _: PathBuf = write_file(&temp_dir, "foo", b"hello");

    let (tx, rx) = mpsc::channel();
    let mut watcher =
        RecommendedWatcher::new(move |event| tx.send(event).unwrap(), Config::default()).unwrap();

    println!("watching {}", temp_dir.path().display());
    watcher
        .watch(temp_dir.path(), RecursiveMode::NonRecursive)
        .unwrap();

    // We haven't triggered any events since starting the watcher
    assert!(rx.try_recv().is_err());

    // Write a file to the watched directory.
    // This should trigger an event where one of the paths is
    // "<temp_dir>/bar".
    let expected_path = write_file(&temp_dir, "bar", b"good-bye");

    let ev = rx.recv().unwrap().unwrap();
    println!("{:#?}", ev);

    assert!(
        ev.paths.contains(&expected_path),
        "expected event for {}, but got event for {:?}?!",
        expected_path.display(),
        ev.paths,
    );

    // Note: as expected, writing the second file triggers a single event:
    //   * with the inotify backend, the event is associated with the newly created file "bar",
    //     as expected
    //   * with the kqueue backend, the event is wrongly associated with file "foo"
}
```
with `Cargo.toml`:

```toml
[package]
name = "notify-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
notify = "6.1.1"
temp-dir = "0.1.14"
```

The assertion fails on platforms using the `kqueue` backend. With the `inotify` backend, it works as expected. 

#### What you expected

```
watching /tmp/tdb9d-0
Event {
    kind: Create(
        File,
    ),
    paths: [
        "<TEMP_DIR_PATH>/bar",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
```

#### What happened
The assertion failed
```
watching /tmp/tb59-0
Event {
    kind: Create(
        File,
    ),
    paths: [
        "/tmp/tb59-0/foo",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
thread main panicked at src/main.rs:38:5:
expected event for /tmp/tb59-0/bar, but got event for ["/tmp/tb59-0/foo"]?!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

------

I think the bug is in [`handle_kqueue`](https://github.com/notify-rs/notify/blob/146b1bd5794e4bd53b112863283f434f63b9a953/notify/src/kqueue.rs#L141)'s handling of `kqueue::Vnode::Write`:
  * it assumes that all the files from the watched directory are already in `self.watches`  (which isn't the case for non-recursive watchers)
  * it then looks for [the first entry from `std::fs::read_dir(&path)`](https://github.com/notify-rs/notify/blob/146b1bd5794e4bd53b112863283f434f63b9a953/notify/src/kqueue.rs#L174-L176) that is not in `self.watches`. However, if the watcher is non-recursive, this entry won't necessarily be the one that triggered the `kqueue::Vnode::Write` event 

I can try to come up with a patch to fix this, if you'd like.

JohnTitor:
## Description

Fix #644

As far as I could tell, and as stated in the issue, there's no accurate information to infer the newly created paths perfectly, it'd be better to tell ambiguious event not to make confusion.

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
 - CHANGE: preserve watched path representation in `Event.paths` and `Watcher::watched_paths`; relative watch paths now produce relative event paths consistently across backends [#453] [#740]
+- FIX: [kqueue] stop reporting arbitrary existing child paths for non-recursive directory write events [#644]
 
 ## notify 9.0.0-rc.3 (2026-04-16)
 
@@ -20,6 +21,7 @@
 [#739]: https://github.com/notify-rs/notify/issues/739
 [#453]: https://github.com/notify-rs/notify/issues/453
 [#740]: https://github.com/notify-rs/notify/issues/740
+[#644]: https://github.com/notify-rs/notify/issues/644
 
 ## notify 9.0.0-rc.2 (2026-02-14)
 
--- a/notify/src/kqueue.rs
+++ b/notify/src/kqueue.rs
@@ -200,9 +200,20 @@ impl EventLoop {
                             Ok(Event::new(EventKind::Remove(RemoveKind::Any)).add_path(event_path))
                         }
 
-                        // a write to a directory means that a new file was created in it, let's
-                        // figure out which file this was
-                        kqueue::Vnode::Write if path.is_dir() => {
+                        // A write to a recursively watched directory may mean that a new file
+                        // was created in it. Non-recursive directory watches do not track
+                        // children, so guessing from read_dir would be unreliable.
+                        // FIXME: harden guessing for non-recursive watches.
+                        // Context: https://github.com/notify-rs/notify/issues/644
+                        kqueue::Vnode::Write
+                            if watch.is_some_and(|watch| watch.is_recursive)
+                                && if self.follow_symlinks {
+                                    path.is_dir()
+                                } else {
+                                    std::fs::symlink_metadata(&path)
+                                        .is_ok_and(|metadata| metadata.is_dir())
+                                } =>
+                        {
                             // find which file is new in the directory by comparing it with our
                             // list of known watches
                             std::fs::read_dir(&path)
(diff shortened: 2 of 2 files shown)


dfaust:
Looks good.

PR #913, opened 2026-04-30
perf: Optimize debouncer.watch() lookup time

### LINKED ISSUE #912: [Bug]: notify_debouncer_full .watch is very slow
### Prerequisites

### Component

notify-debouncer-full

### Operating System

Linux

### Notify Version

latest main

### Notify Backend

Unknown/Auto-detected

## What happened?

This line of code 

https://github.com/notify-rs/notify/blob/62fb95385a9821e96f534f3cd20f87b33dae73d4/notify-debouncer-full/src/lib.rs#L590

does a sequential scan for every call, if it is called in the iterator it will do an always growing linear scan 

this is how it compares to the raw notify watch 

```
┌────────────────────────────────────────────┬────────────────────────┬───────────────────────┐ 
│Backend                                     │30k dirs                │Per call avg           │ 
├────────────────────────────────────────────┼────────────────────────┼───────────────────────┤ 
│Raw notify::Watcher::watch()                │211 ms                  │7 µs                   │ 
├────────────────────────────────────────────┼────────────────────────┼───────────────────────┤ 
│debouncer.watch()                           │10.9 s                  │364 µs                 │ 
└────────────────────────────────────────────┴────────────────────────┴───────────────────────┘ 

```

## Steps to reproduce

call debouncer.watch in the loop

dmtrKovalenko:
## Changelog

Optimize performance of debouncer.watch() calls

## Testing
tested 

## Description
Use sorted VecDeque on 30k invocations cuts down all the linear scanning work that has be done before. Cuts down 10s to 300ms

## Related Issues

closes https://github.com/notify-rs/notify/issues/912

### DIFF
--- a/notify-debouncer-full/src/lib.rs
+++ b/notify-debouncer-full/src/lib.rs
@@ -355,17 +355,19 @@ impl<T: FileIdCache> DebounceDataInner<T> {
         }
     }
 
-    fn recursive_mode(&mut self, path: &Path) -> RecursiveMode {
-        self.roots
-            .iter()
-            .find_map(|(root, recursive_mode)| {
-                if path.starts_with(root) {
-                    Some(*recursive_mode)
-                } else {
-                    None
+    fn recursive_mode(&self, path: &Path) -> RecursiveMode {
+        for ancestor in path.ancestors() {
+            if let Ok(index) = self
+                .roots
+                .binary_search_by(|(root, _)| root.as_path().cmp(ancestor))
+            {
+                if self.roots[index].1 == RecursiveMode::Recursive {
+                    return RecursiveMode::Recursive;
                 }
-            })
-            .unwrap_or(RecursiveMode::NonRecursive)
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


notify-debouncer-full/src/lib.rs
```diff
@@ -586,13 +589,17 @@ impl<T: Watcher, C: FileIdCache> Debouncer<T, C> {
 
         let mut data = self.data.lock().unwrap();
 
-        // skip, if the root has already been added
-        if data.roots.iter().any(|(p, _)| p == &path) {
-            return;
+        match data
+            .roots
+            .binary_search_by(|(p, _)| p.as_path().cmp(path.as_path()))
```
dmtrKovalenko, line 596: the path comparison is actually pretty slow in rust as well we could've compared underlying os_str but this will not work with non canonicalized paths

if we do that - performance impact would be absolutely unnoticable

JohnTitor (approved):
Thank you!

PR #919, opened 2026-05-08
Add freebsd_inotify feature to use inotify on FreeBSD 14.4+

kev009:
## Description
FreeBSD 14.4+ provides an in-kernel inotify(2) implementation and libc wrappers in libc.so.7. Switching FreeBSD's notify backend from kqueue to inotify is a large win on big trees because kqueue requires one open file descriptor per watched directory (or file) while inotify uses a single fd with watch descriptors, scaling to large repositories without hitting RLIMIT_NOFILE / kern.maxfiles_per_proc.

Add a `freebsd_inotify` cargo feature (default off). When enabled:

  * `RecommendedWatcher` resolves to `INotifyWatcher` on FreeBSD.
  * The `inotify` module compiles in (it uses only std::os::unix + libc fcntl flags, no Linux-specific syscalls).
  * The `inotify` crate is pulled in via an optional FreeBSD target dep, and `inotify-sys/freebsd-native` is forwarded to skip linking the userspace libinotify shim.
  * The `BoundSender` / `bounded` helpers used by the inotify backend are extended to compile under this cfg path.

When the feature is off, FreeBSD continues to use the kqueue backend, so this is a strict opt-in and pre-14.4 FreeBSD remains buildable.

Two existing inotify-backend tests encoded Linux-specific event semantics; both are adjusted here so the test suite is green on FreeBSD:

  * `set_file_mtime` widened from `modify_data_any()` to `modify()`. Linux's inotify reports utimensat(2) as IN_MODIFY (Data); FreeBSD's reports it as IN_ATTRIB (Metadata), matching inotify(7)'s own description (which lists timestamps under IN_ATTRIB). Both still satisfy the portable contract "mtime change fires a Modify event".

  * `write_to_a_hardlink_…doesnt_trigger…` gated to #[cfg(target_os = "linux")] and renamed with an `_on_linux` suffix. Linux's inotify is dirent-centric (events fire on the pathname used for the write); FreeBSD's is inode-centric (events fire on every pathname pointing to the modified inode). The test as written encodes Linux's behavior; a comment documents the underlying architectural difference.

Tested on FreeBSD: cargo test --workspace --features notify/freebsd_inotify reports 74 passed, 0 failed, 2 ignored in `notify` lib unittests, with all other workspace crates fully green. Linux behavior is unchanged.

### DIFF
--- a/notify/Cargo.toml
+++ b/notify/Cargo.toml
@@ -22,6 +22,11 @@ default = ["macos_fsevent"]
 serde = ["notify-types/serde"]
 macos_kqueue = ["kqueue", "mio"]
 macos_fsevent = ["objc2-core-foundation", "objc2-core-services"]
+# Use inotify (instead of kqueue) as the FreeBSD backend. Requires FreeBSD
+# 14.4+, which provides inotify(2) and libc wrappers in the base system.
+# Forwards `freebsd-native` to `inotify-sys` so it resolves the inotify
+# symbols against libc rather than the userspace `libinotify` shim.
+freebsd_inotify = ["inotify", "inotify/freebsd-native"]
 serialization-compat-6 = ["notify-types/serialization-compat-6"]
 tokio = ["dep:tokio"]
 futures = ["dep:futures"]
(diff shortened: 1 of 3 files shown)


kev009:
Marked as a draft for now, this needs stable releases of https://github.com/hannobraun/inotify-rs/pull/258

PR #923, opened 2026-05-11
fix(windows): emit Remove event when watched directory is deleted

rtfeldman:
## Description

When a watched directory is deleted on Windows, no `Remove` event is currently emitted even though it would be on macOS and Linux. This is because [`ReadDirectoryChangesW`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesw) returns `ERROR_ACCESS_DENIED` when a watched directory is deleted, and the existing handler (PR #674) silently unwatches (without emitting any event to the consumer).

This PR makes the Windows backend match the behavior of inotify (`IN_DELETE_SELF` from PR #540) and FSEvents (`ROOT_CHANGED` + `ITEM_REMOVED`), which both emit a `Remove` event when their watched root disappears. The new Windows event is only emitted for directory watches the user explicitly asked for; it excludes file watches because `FILE_ACTION_REMOVED` already fires when a non-directory file is deleted. This also adds a new `delete_self_dir` test for each backend's test module.

## Related Issues

None, but I can open an issue to report the current discrepancy if it would help!

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+- FIX: [windows] emit a Remove event when a watched directory is deleted, matching inotify and FSEvents
+
 ## notify 9.0.0-rc.4 (2026-05-02)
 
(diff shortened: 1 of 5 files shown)


JohnTitor (approved):
Thanks!

PR #930, opened 2026-05-21
feat(notify): make FSEvents stream latency configurable

bvinc:
Adds Config::with_fsevent_latency to control the latency passed to FSEventStreamCreate. Defaults to Duration::ZERO, preserving current behavior. Set at construction time only.

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## unreleased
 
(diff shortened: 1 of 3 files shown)


bvinc:
This seems to be important as it's basically the only knob you can turn to hopefully reduce missed events on mac fsevent.  So this PR allows it to be configured.

JohnTitor (approved):
Thanks!

PR #933, opened 2026-05-29
feat(notify-debouncer-full): use condvar for no idle CPU usage

Pandapip1:
## Description

The existing code spawned a thread that looped forever and slept for one tick. This is valid, but results in some idle CPU usage.

This adds a condvar that means that the debouncer loop will go back to sleep until such time as there are actually events to handle, eliminating the idle CPU usage.

## Related Issues

None

### DIFF
--- a/notify-debouncer-full/src/lib.rs
+++ b/notify-debouncer-full/src/lib.rs
@@ -172,7 +172,13 @@ impl DebounceEventHandler for std::sync::mpsc::Sender<DebounceEventResult> {
 /// Comes with either a vec of events or vec of errors.
 pub type DebounceEventResult = Result<Vec<DebouncedEvent>, Vec<Error>>;
 
-type DebounceData<T> = Arc<(Mutex<DebounceDataInner<T>>, Condvar)>;
+type DebounceData<T> = Arc<SharedDebounceData<T>>;
+
+#[derive(Debug)]
+struct SharedDebounceData<T> {
+    inner: Mutex<DebounceDataInner<T>>,
+    changed: Condvar,
+}
 
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 struct Queue {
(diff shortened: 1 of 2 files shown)


Pandapip1:
Fixed formatting.

Pandapip1:
I think there must have been a race in the previous version. Simplified it significantly, and this one *shouldn't* have a race.

notify-debouncer-full/src/lib.rs
```diff
@@ -578,6 +577,7 @@ impl<T: Watcher, C: FileIdCache> Debouncer<T, C> {
 
     fn set_stop(&self) {
         self.stop.store(true, Ordering::Relaxed);
+        self.data.1.notify_all();
```
JohnTitor: I guess this notification is not synchronized well with the mutex used by the condvar wait, right? If I recognize it correctly, if `stop()` runs after the worker checks the empty predicate but before it actually parks in `Condvar::wait`, this notify can be lost and `stop()` can hang forever on `join()`. Maybe we have to coordinate them better.

notify-debouncer-full/src/lib.rs
```diff
@@ -759,17 +762,19 @@ pub fn new_debouncer_opt<F: DebounceEventHandler, T: Watcher, C: FileIdCache + S
     let thread = std::thread::Builder::new()
         .name("notify-rs debouncer loop".to_string())
         .spawn(move || loop {
-            if stop_c.load(Ordering::Acquire) {
-                break;
+            let mut lock = data_c.0.lock().unwrap();
+            if lock.queues.is_empty() && lock.errors.is_empty() && lock.rename_event.is_none() && lock.rescan_event.is_none() {
```
JohnTitor: The condition looks buggy, `rename_event` is matching state and is not drained by `debounced_events()`.

Pandapip1:
I *believe* this addresses both comments; LMK if there's anything specific I should change.

JohnTitor (approved):
Thank you!

PR #934, opened 2026-05-31
chore(deps): bump kqueue to 1.2.0

worr:
## Description

This version adds a number of fixes to major problems, specifically in that it eliminates panics in failure conditions and substantially increases performance of watch removals.


JohnTitor (approved):
Thanks!

PR #943, opened 2026-06-19
Refactored `notify-debouncer-mini`

cactusdualcore:
## Description
I made some stylistic changes that do not increase the MSRV above the current one, but in my opinion improve the overall structure of the code.

### DIFF
--- a/notify-debouncer-mini/src/lib.rs
+++ b/notify-debouncer-mini/src/lib.rs
@@ -158,39 +158,30 @@ where
     }
 }
 
-#[cfg(feature = "crossbeam-channel")]
-impl DebounceEventHandler for crossbeam_channel::Sender<DebounceEventResult> {
-    fn handle_event(&mut self, event: DebounceEventResult) {
-        let _ = self.send(event);
-    }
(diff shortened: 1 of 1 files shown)


JohnTitor:
Thanks for the PR! But I don't think this is a change we should merge right now. This is mostly an internal/stylistic refactor, and the current code is small and fairly explicit.

I'm going to close this for now. If there's a specific bug or behavior change you wanted to enable, a smaller focused PR for that would be easier to evaluate.

That said, I appreciate the contribution, thanks again.

cactusdualcore:
I think the current API of `notify` is somewhat difficult to work with. This is to be a bit expected, given it feels like a "low level" crate. I assume this is why the debouncer crates even exist (besides being usage examples).

But imho, their APIs are not really any better. My idea was to improve both their APIs and increase the consistency between them. I just didn't want to go ham at it and rework an entire public API out of the blue though. So I tried to first test the waters with this PR.

@JohnTitor If you do not mind, we can discuss in more detail what I have in mind for the new API. I am willing to implement it myself. Of course only if you are interested. Otherwise I'll just write my own debouncer crate on top of notify.

PR #950, opened 2026-07-06
Add `watch_filtered`, filtering paths to watch

daandemeyer:
Adds `WatchFilter` and `Watcher::watch_filtered`, allowing directories to be excluded when setting up a watch. Implemented for the inotify, fsevents, kqueue, windows, poll, and null backends, with filter support plumbed through `Watcher::update_paths`, `notify-debouncer-full`, and the `FileIdCache` API.

The filter gates directories only: walk-based backends prune excluded directories at scan time and gate directories discovered while watching; FSEvents and Windows suppress matching events at delivery time. File watches are never affected.

Two restrictions keep the semantics simple and the implementation small:

- Watching a directory the filter itself rejects returns the new `ErrorKind::PathExcluded` and changes nothing. Symlink roots are checked against their resolved target too.
- A directory watch carrying a filter must not overlap another directory watch in either direction; such calls are refused with an error. Filters are never merged across watches, so each watch entry carries exactly one filter. Accept-all watches keep the existing overlap semantics, file watches never conflict, and re-watching the same path still replaces the watch (`WatchFilter::same_filter` identity makes an identical re-watch a no-op).

Based on the filter-watch-targets branch:
- Add `watch_filtered`, filtering paths to watch.
- Add fsevent support and filter newly added paths too.
- Add support for watch_filtered to Windows.
- Add filtered watch implementation for kqueue.

### DIFF
--- a/docs/UPGRADING_V8_TO_V9.md
+++ b/docs/UPGRADING_V8_TO_V9.md
@@ -104,6 +104,57 @@ watcher.watch(path, notify::RecursiveMode::NonRecursive)?;
 watcher.unwatch(path)?;
 ```
 
+### 5) `Watcher::watch_filtered` is the required trait method
+
+v9 adds `Watcher::watch_filtered(path, recursive_mode, watch_filter)`, which excludes
+directories rejected by a `WatchFilter` from the watch. It is now the required method of the
+`Watcher` trait; `Watcher::watch` is a provided method that forwards to it with
+`WatchFilter::accept_all()`.
+
+Users of `Watcher` are unaffected: `watch` keeps working unchanged. Implementors of the trait
+must rename their `watch` implementation to `watch_filtered` and accept the extra parameter.
+
+The filter gates directories only: watching a non-directory path must never be affected by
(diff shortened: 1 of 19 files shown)


daandemeyer:
@JohnTitor Any chance you could enable workflows for this PR? The workflows don't run on my fork so I need them enabled here to make sure windows and macos work as expected.

JohnTitor:
Triggered!

daandemeyer:
@JohnTitor Hmm seems you need to reapprove every single time. I'll see if I can find some trivial fix that can already land so I don't have to ask every time

daandemeyer:
Opened https://github.com/notify-rs/notify/pull/951, if you merge that one CI will start working automatically here without needing your approval every time.

daandemeyer:
Should be ready for review now, includes two other commits which I've split out into separate PRs. Will rebase this pr once those are merged. It's a lot of code, but a lot of it is tests, happy to split up into separate commits or PRs if that's preferred.

daandemeyer:
Closing so I can open other prs

PR #952, opened 2026-07-08
feat(notify): use inotify as the FreeBSD backend

kev009:
## Description
    FreeBSD 15.0+ provides in-kernel inotify(2) and libc wrappers. Enable
    the freebsd_inotify feature by default so INotifyWatcher is
    RecommendedWatcher on FreeBSD (same as Linux/Android), keep kqueue
    available, and bump inotify to 0.11.4.
    
    FreeBSD 14.x has no inotify in base: build with default-features = false
    to use kqueue. Native linkage is handled in inotify-sys 0.1.8.
    
    Test adjustments for FreeBSD event semantics:
    - set_file_mtime accepts any Modify (Linux: IN_MODIFY, FreeBSD: IN_ATTRIB)
    - hardlink write-through test is Linux-only (dirent- vs inode-centric)
    
    CI: FreeBSD 15.1 with defaults; FreeBSD 14.4 with --no-default-features.
    Document the platform split in README/changelog.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -63,9 +63,9 @@ body:
       description: Which backend is being used?
       options:
         - Unknown/Auto-detected
-        - inotify (Linux)
+        - inotify (Linux/Android/FreeBSD 15+)
         - fsevents (macOS)
-        - kqueue (macOS/BSD)
+        - kqueue (macOS/BSD/FreeBSD 14.x)
         - ReadDirectoryChangesW (Windows)
         - polling (all platforms)
     validations:
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -64,8 +64,8 @@ body:
       options:
         - Unknown/Auto-detected
         - inotify (Linux/Android/FreeBSD 15+)
-        - fsevents (macOS)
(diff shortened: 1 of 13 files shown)


kev009:
Unfortunately inotify support was not added in 14.4.

I added a matrix build to deal with this, we prefer inotify on 15.0+ and offer an opt out for someone stuck on 14.x.

JohnTitor:
Defaulting to inotify is too rush to me, given the FreeBSD 14 's EoL is far enough (Nov. 2028). Let's stay on kqueue by default and opt-in inotify with a feature flag.

JohnTitor (approved):
Thanks!

kev009:
@JohnTitor many thanks.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=296835 prompted me to do this.  for whatever reason kqueue is not happy with common consumers of this crate (Zed, codex, grok for example) where it will hang and the crash dump can take a very long time and is uninterruptible until this fix which is not in 14 or 15 yet.

Is there a way to make Cargo change preference at **build** time?  It would fit FreeBSD well to use inotify when it detects FreeBSD 15 or newer, and the way binaries are built on FreeBSD this is a safe decision.

JohnTitor:
Hm, it sounds worth addressing then. We cannot do anything before dependency solution but could add a FreeBSD version detector and switch the default by it (n<15 or 15<=n). Let me add such a tweak.

PR #953, opened 2026-07-09
kqueue: preserve the error kind returned through the event-loop channel

daandemeyer:
watch_inner and unwatch_inner applied `map_err(|e| Error::generic(&e.to_string()))`
to the inner add_watch / remove_watch result, collapsing every error kind
(PathNotFound, Io, ...) into Generic and dropping the attached
path. Map only the channel recv failure instead and propagate the inner result
unchanged, so callers observe the real error kind; this also removes the panic
when the event loop has already shut down.


JohnTitor (approved):
Makes sense!

PR #954, opened 2026-07-09
Move orphaned integration tests into notify/tests

daandemeyer:
race-with-remove-dir, poll-watcher-hashing, and serialise-events lived at
the workspace root, which is a virtual manifest with no package, so cargo
never compiled or ran them. Move them under notify/tests/ so they run again
as part of `cargo test`. Also remove the defunct and-retry helper and its
.gitignore entries: a Travis wrapper that reran failed cargo tests, orphaned
since the switch to GitHub Actions.

Restoring the moved tests required catching up with API drift:

- poll-watcher-hashing: port PollWatcherConfig/with_config to
  Config/PollWatcher::new and futimens to its AsFd signature; enable the
  nix dev-dependency's fs and time features.
- serialise-events: update the serde expectations to the current
  internally-tagged ({"type": ...}) event format.


JohnTitor (approved):
Thanks!

PR #957, opened 2026-07-13
kqueue: install recursive watches before delivering events

daandemeyer:
Kqueue queued watches for newly discovered entries but invoked the event
handler before applying them. If a handler reacted to a directory-create
event by modifying that directory, the change could happen before its
recursive watch was installed and be lost permanently.

Buffer events until the queued watch removals and additions have been
applied. Delivering a directory-create event now guarantees that recursive
coverage for the new directory is already active.

Add a regression test whose handler creates a file in the newly reported
directory and verifies that the resulting event is observed.

### DIFF
--- a/notify/src/kqueue.rs
+++ b/notify/src/kqueue.rs
@@ -174,6 +174,7 @@ impl EventLoop {
     fn handle_kqueue(&mut self) {
         let mut add_watches = Vec::new();
         let mut remove_watches = Vec::new();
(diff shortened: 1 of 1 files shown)


daandemeyer:
Gentle ping, this fixes a CI failure on my other PR so would love to get this merged so I get green CI on the other one.

JohnTitor (approved):
Sorry for the delay, LGTM

PR #958, opened 2026-07-13
windows: handle watch shutdown races

daandemeyer:
ReadDirectoryChangesW can complete an outstanding request just before
CancelIo runs. When stop_watch closed the directory handle and entered its
alertable wait, that already-queued callback still arrived with
ERROR_SUCCESS and tried to rearm the request using the closed handle. The
resulting ERROR_INVALID_HANDLE was emitted to users and caused overlapping
watch tests to fail intermittently.

Share a stopping flag between each watch state and its completion requests.
Set it before cancellation begins, then have callbacks release the completion
semaphore and return without processing or rearming once the watch is
stopping. This preserves the existing wait-for-completion lifetime guarantee
without submitting I/O against a closed handle.

Directory deletion has a second race: RemoveDirectory can return while the
open watch handle keeps the directory delete-pending. In that window,
try_exists may report an access error instead of absence, causing the callback
to rearm the invalidated handle and emit ERROR_ACCESS_DENIED. Query
FILE_STANDARD_INFO on the existing handle and treat DeletePending as directory
removal while preserving genuine permission failures.

The stopped-completion regression test verifies that a queued successful
completion is acknowledged without an error or another unwatch. The existing
delete_self_dir test covers removal notification for delete-pending directory
handles.

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -5,12 +5,14 @@
 - FEATURE: [macOS] add `Config::with_fsevent_latency` to configure FSEvents stream latency [#930]
 - FIX: [windows] emit a Remove event when a watched directory is deleted, matching inotify and FSEvents
 - FIX: [windows] surface `ReadDirectoryChangesW` read-start failures [#935]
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Thanks!

PR #960, opened 2026-07-15
inotify: coalesce kernel-driven watch cleanup

daandemeyer:
Linux removes watch descriptors automatically when a watched inode is
deleted and emits DELETE_SELF and IGNORED. The event loop previously
queued the same path multiple times, so cleanup called inotify_rm_watch
on an invalid descriptor and then reported WatchNotFound for the
duplicate removal.

Coalesce pending removals by path and record whether the kernel already
removed the root descriptor. Process removals deepest-first, handle
IGNORED even when remove events are filtered, and remove potentially
live descendant descriptors separately. Treat expected EINVAL deletion
races as trace output while warning on unexpected failures.

Add regression coverage for IGNORED-only cleanup, duplicate removal
coalescing, and moved-out descendant descriptor cleanup.

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
 - FEATURE: [macOS] add `Config::with_fsevent_latency` to configure FSEvents stream latency [#930]
+- FIX: [linux] coalesce duplicate inotify cleanup events and handle kernel-removed descriptors without spurious `EINVAL` or `WatchNotFound` logs
 - FIX: [windows] emit a Remove event when a watched directory is deleted, matching inotify and FSEvents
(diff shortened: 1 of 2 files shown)


daandemeyer:
@JohnTitor I have lots of other improvements queued up but can't submit them for testing due to the PR limit on the repo. Could you please bump that PR limit so I don't have to baby sit my PRs all the time?

JohnTitor:
The limit is to prevent the flood of notifications/slops for maintainers and I cannot bump it. Please understand we all are unpaid OSS maintainers and reviewing takes some time. I need firstly make enough money to live on in work, and I do not have enough financial resources to commit OSS full-time. Thank you for your patience.

JohnTitor (approved):
LGTM, thanks!

PR #964, opened 2026-07-21
fix(windows): emit rescan events for lost change details

### LINKED ISSUE #963: [Bug]: Windows ReadDirectoryChangesW overflow does not emit a rescan event
### Prerequisites

### Component

notify (core library)

### Operating System

Windows

### Notify Version

notify = "8.2.0"

### Notify Backend

ReadDirectoryChangesW (Windows)

## What happened?

Hi! I ran into this while deciding how to use notify in a local editor. I need to keep an in-memory project tree correct, so `Event::need_rescan()` is the important escape hatch whenever the native event details are lost.

Related context exists in
- [#89](https://github.com/notify-rs/notify/issues/89),
- [#206](https://github.com/notify-rs/notify/issues/206),
- [#412](https://github.com/notify-rs/notify/issues/412), and
- [#434](https://github.com/notify-rs/notify/issues/434), but I could not find a focused open issue for the zero-byte Windows completion being ignored. In particular, #412 shows missed events under load but does not prove this exact overflow path.

The Windows completion callback receives the number of transferred bytes as `_bytes_written`, but does not inspect it. `ERROR_SUCCESS` always falls through to the `FILE_NOTIFY_INFORMATION` parser:

- [`handle_event` ignores `_bytes_written`](https://github.com/notify-rs/notify/blob/a1d7c2d8f80786679d58ec6d5986a1d4278bc8cf/notify/src/windows.rs#L331-L385)
- [the parser then reads the request buffer](https://github.com/notify-rs/notify/blob/a1d7c2d8f80786679d58ec6d5986a1d4278bc8cf/notify/src/windows.rs#L378-L463)

Microsoft documents that when the `ReadDirectoryChangesW` buffer overflows, the buffer contents are discarded and the transferred-byte count is zero. It says the caller should enumerate the directory/subtree to recover current state. The same documentation says `ERROR_NOTIFY_ENUM_DIR` means the system could not record all changes and also requires enumeration:

- [`ReadDirectoryChangesW` documentation](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesw)

So on a successful zero-byte completion, notify currently has the exact signal that details were lost but does not turn it into `Flag::Rescan`. With the zeroed request buffer there is normally no recognized file action to emit, so a caller can miss changes without receiving the API's recovery signal.

I think the minimal expected behavior is:

1. If `error_code == ERROR_SUCCESS` and `bytes_written == 0`, do not parse the    buffer.
2. Rearm the watch and emit one pathless `EventKind::Other` event with `Flag::Rescan` so `event.need_rescan()` returns true.
3. Treat `ERROR_NOTIFY_ENUM_DIR` as the same loss-of-detail condition wherever the backend can receive it.
4. Let the application enumerate. I do not think notify itself should recursively rescan the user's directory or invent per-path events.

The .NET Windows watcher handles this same native contract by treating `numBytes == 0` as an internal-buffer overflow and parsing only the reported byte extent:

- [`.NET FileSystemWatcher.Windows.cs`](https://github.com/dotnet/runtime/blob/6c8b0c7c833028bad09fbcaff54ac3b42df5c3d3/src/libraries/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Windows.cs#L234-L280)

There is a related but broader problem in #206: several native/rearm errors are logged or dropped instead of reaching the event handler. I am happy to keep that as a separate change if combining it would make this fix harder to review. The overflow case itself already maps directly to notify's existing rescan API.

## Steps to reproduce

There are two levels of reproduction.

The deterministic source-level reproduction is:

1. Open `notify/src/windows.rs` at the 8.2.0 tag.
2. Follow the completion routine's `bytes_written` argument.
3. Observe that it is named `_bytes_written` and never read.
4. Follow `ERROR_SUCCESS`; it rearms and parses the buffer regardless of a zero transferred count.
5. Compare that with the Microsoft overflow contract linked above.

For a live run, watch one temporary directory recursively, deliberately slow the event callback, and create many short-named files from several threads. Log both the native completion count in a temporary instrumented build and every `event.need_rescan()` result. On the affected code, a zero-byte completion can be observed without a corresponding rescan event. Exact file/thread counts need tuning for the machine, so I would not use a timing-only stress run as the regression test.

## Environment details

- Windows: **Windows 11 Pro 25H2, build 26200.8875**
- Filesystem: **local NTFS fixed drive (`C:`), not SMB**
- Rust: rustc 1.95.0 (59807616e 2026-04-14)
- Watch mode: one recursive directory
- Buffer in notify Windows backend: 16 KiB

## Error messages or logs

There is usually no public error for the zero-byte success path. That is the problem: it enters the success branch and no `need_rescan()` event reaches the consumer.

If `ERROR_NOTIFY_ENUM_DIR` or another unexpected code reaches the callback, the current fallback logs an error containing the directory and unwatches it rather than returning a typed error to the handler. A synchronous `ReadDirectoryChangesW` submission failure is also released without a handler event in [`start_read`](https://github.com/notify-rs/notify/blob/a1d7c2d8f80786679d58ec6d5986a1d4278bc8cf/notify/src/windows.rs#L270-L327). That broader delivery behavior overlaps #206.

## Minimal code example

This is a small observer/stress harness, not a deterministic regression test. Add `tempfile` as a dev-dependency, run it on Windows, and tune `WRITERS`, `FILES`, or the callback delay until an instrumented backend reports a zero-byte completion.

```rust
use notify::{Watcher, RecommendedWatcher, Result};
use notify::{Event, RecursiveMode};
use std::fs;
use std::sync::mpsc;
use std::thread;
use std::time::{Duration, Instant};

const WRITERS: usize = 8;
const FILES_PER_WRITER: usize = 10_000;

fn main() -> Result<()> {
    let root = tempfile::tempdir()?;
    let (tx, rx) = mpsc::channel::<bool>();

    let mut watcher: RecommendedWatcher =
        notify::recommended_watcher(move |result: notify::Result<Event>| {
            // Make it easier for the fixed native buffer to fill before the Windows
            // watcher thread can return to its alertable wait.
            thread::sleep(Duration::from_millis(10));
            let needs_rescan = result.as_ref().is_ok_and(Event::need_rescan);
            let _ = tx.send(needs_rescan);
        })?;
    watcher.watch(root.path(), RecursiveMode::Recursive)?;

    let mut writers = Vec::new();
    for writer in 0..WRITERS {
        let path = root.path().to_path_buf();
        writers.push(thread::spawn(move || -> std::io::Result<()> {
            for file in 0..FILES_PER_WRITER {
                fs::write(path.join(format!("{writer}-{file}.tmp")), b"x")?;
            }
            Ok(())
        }));
    }

    for writer in writers {
        writer.join().expect("writer thread panicked")?;
    }

    let deadline = Instant::now() + Duration::from_secs(10);
    let mut events = 0usize;
    let mut rescans = 0usize;
    while Instant::now() < deadline {
        match rx.recv_timeout(Duration::from_millis(100)) {
            Ok(needs_rescan) => {
                events += 1;
                if needs_rescan {
                    rescans += 1;
                }
            }
            Err(mpsc::RecvTimeoutError::Timeout) => {}
            Err(mpsc::RecvTimeoutError::Disconnected) => break,
        }
    }

    println!("events={events}, rescans={rescans}");
    Ok(())
}
```

For the actual regression test, I would inject the completion tuple directly so the test proves `ERROR_SUCCESS + 0 bytes -> need_rescan()` without relying on a particular machine being slow enough to overflow.

srothgan:
## Description

The bug reported in #963 occurs when Windows discards change details reported through `ReadDirectoryChangesW`. Its completion routine signals this in two ways: a successful completion with zero transferred bytes, or `ERROR_NOTIFY_ENUM_DIR`.

The backend previously ignored the transferred-byte count and continued into the normal buffer parser. For a zero-byte completion, this meant consumers received no indication that their view of the filesystem might be stale.

This change deliberately keeps the fix small:

- rearm the watch before handling the completion
- emit one pathless `EventKind::Other` event with `Flag::Rescan`
- skip the discarded notification buffer
- preserve the existing record parser and unrelated error handling
- preserve the existing rearm error if rearming also fails

The regression test covers both lost-details signals, verifies the pathless rescan event, and confirms that ordinary successful and cancelled completions are not treated as rescans.

I tested this on Windows with:

- `cargo test -p notify --lib` (74 passed)
- `cargo check --workspace --all-targets --all-features`
- `cargo fmt --all -- --check`

## Related Issues

- Fixes #963

### DIFF
--- a/notify/src/windows.rs
+++ b/notify/src/windows.rs
@@ -21,8 +21,8 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use windows_sys::Win32::Foundation::{
-    CloseHandle, ERROR_ACCESS_DENIED, ERROR_OPERATION_ABORTED, ERROR_SUCCESS, HANDLE,
-    INVALID_HANDLE_VALUE, WAIT_OBJECT_0,
+    CloseHandle, ERROR_ACCESS_DENIED, ERROR_NOTIFY_ENUM_DIR, ERROR_OPERATION_ABORTED,
+    ERROR_SUCCESS, HANDLE, INVALID_HANDLE_VALUE, WAIT_OBJECT_0,
 };
 use windows_sys::Win32::Storage::FileSystem::{
     CreateFileW, FileStandardInfo, GetFileInformationByHandleEx, ReadDirectoryChangesW,
@@ -517,9 +517,16 @@ fn is_delete_pending(handle: HANDLE) -> bool {
     success != 0 && info.DeletePending
 }
 
+// Windows reports discarded notification details as either a zero-byte successful completion or
+// ERROR_NOTIFY_ENUM_DIR.
+fn completion_rescan_event(error_code: u32, bytes_written: u32) -> Option<Event> {
+    (error_code == ERROR_NOTIFY_ENUM_DIR || (error_code == ERROR_SUCCESS && bytes_written == 0))
+        .then(|| Event::new(EventKind::Other).set_flag(Flag::Rescan))
+}
+
 unsafe extern "system" fn handle_event(
     error_code: u32,
-    _bytes_written: u32,
+    bytes_written: u32,
     overlapped: *mut OVERLAPPED,
 ) {
     let overlapped: Box<OVERLAPPED> = Box::from_raw(overlapped);
@@ -573,8 +580,8 @@ unsafe extern "system" fn handle_event(
                 return;
             }
         }
-        ERROR_SUCCESS => {
-            // Success, continue to handle the event
+        ERROR_SUCCESS | ERROR_NOTIFY_ENUM_DIR => {
+            // Continue below to rearm the watch before handling the completion.
         }
         _ => {
             // Some unidentified error occurred, log and unwatch the directory, then return.
@@ -598,6 +605,18 @@ unsafe extern "system" fn handle_event(
         request.action_tx.clone(),
     )
     .err();
+
+    if let Some(event) = completion_rescan_event(error_code, bytes_written) {
+        emit_event(&request.event_handler, Ok(event));
+
+        // Lost event details and a failed rearm are separate conditions. Report both so callers
+        // know to rebuild their state and that this watch is no longer active.
+        if let Some(err) = rearm_error {
+            emit_event(&request.event_handler, Err(err));
+            request.unwatch();
+        }
+        return;
+    }
 
     // The FILE_NOTIFY_INFORMATION struct has a variable length due to the variable length
     // string as its last member. Each struct contains an offset for getting the next entry in
@@ -888,9 +907,12 @@ pub mod tests {
     use std::thread;
     use tempfile::{tempdir, tempdir_in};
 
-    use super::{normalize_path_separators, trim_leading_separators, SeparatorStyle};
+    use super::{
+        completion_rescan_event, normalize_path_separators, trim_leading_separators, SeparatorStyle,
+    };
     use crate::{
-        test::*, ReadDirectoryChangesWatcher, RecursiveMode, Watcher, WindowsPathSeparatorStyle,
+        test::*, Event, EventKind, ReadDirectoryChangesWatcher, RecursiveMode, Watcher,
+        WindowsPathSeparatorStyle,
     };
 
     use std::time::Duration;
@@ -985,6 +1007,22 @@ pub mod tests {
         let path = PathBuf::from(r"\\?\C:\very\long\file");
         let normalized = normalize_path_separators(path, SeparatorStyle::Slash);
         assert_eq!(normalized, PathBuf::from(r"\\?\C:/very/long/file"));
+    }
+
+    #[test]
+    fn lost_change_details_produce_a_pathless_rescan_event() {
+        use windows_sys::Win32::Foundation::{
+            ERROR_NOTIFY_ENUM_DIR, ERROR_OPERATION_ABORTED, ERROR_SUCCESS,
+        };
+
+        let event: Event = completion_rescan_event(ERROR_SUCCESS, 0).expect("rescan event");
+
+        assert_eq!(event.kind, EventKind::Other);
+        assert!(event.paths.is_empty());
+        assert!(event.need_rescan());
+        assert!(completion_rescan_event(ERROR_NOTIFY_ENUM_DIR, 0).is_some());
+        assert!(completion_rescan_event(ERROR_SUCCESS, 12).is_none());
+        assert!(completion_rescan_event(ERROR_OPERATION_ABORTED, 0).is_none());
     }
 
     #[test]
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -6,6 +6,7 @@
 - FIX: [windows] emit a Remove event when a watched directory is deleted, matching inotify and FSEvents
 - FIX: [windows] surface `ReadDirectoryChangesW` read-start failures [#935]
 - FIX: [windows] prevent watch shutdown and watched-directory deletion races from emitting spurious I/O errors [#958]
+- FIX: [windows] emit a rescan event when `ReadDirectoryChangesW` discards change details [#964]
 - FEATURE: [windows] report created file/folder kinds when they can be determined [#935]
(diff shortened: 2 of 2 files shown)


srothgan:
Anything missing from my side before this can be merged? Resolving this fix would help be massively on my own new editor project

notify/src/windows.rs
```diff
@@ -599,113 +700,113 @@ unsafe extern "system" fn handle_event(
     )
     .err();
 
-    // The FILE_NOTIFY_INFORMATION struct has a variable length due to the variable length
-    // string as its last member. Each struct contains an offset for getting the next entry in
-    // the buffer.
-    let mut cur_offset: *const u8 = request.buffer.as_ptr();
-    // In Wine, FILE_NOTIFY_INFORMATION structs are packed placed in the buffer;
-    // they are aligned to 16bit (WCHAR) boundary instead of 32bit required by FILE_NOTIFY_INFORMATION.
-    // Hence, we need to use `read_unaligned` here to avoid UB.
```
dfaust, line 608: What happened to this part? Can the new parser handle 16bit-aligned structs?

notify/src/windows.rs
```diff
@@ -599,113 +700,113 @@ unsafe extern "system" fn handle_event(
...
+                rescan_emitted = true;
+            }
+        },
     }
 
     if let Some(err) = rearm_error {
+        if !rescan_emitted && is_notify_enum_dir_error(&err) {
+            emit_rescan(&request.event_handler);
+        }
         emit_event(&request.event_handler, Err(err));
```
dfaust: Should both events be emitted in case of a re-scan?

notify/src/windows.rs
```diff
@@ -987,6 +1088,253 @@ pub mod tests {
...
+        backing_buffer.extend_from_slice(&[0xff; 32]);
+
+        assert_eq!(
+            super::parse_records(&backing_buffer[..transferred_len]),
+            Ok(vec![super::FileNotifyRecord {
+                action: FILE_ACTION_ADDED,
+                file_name: vec!['a' as u16],
+            }])
+        );
+    }
```
dfaust: This test looks useless.

notify/src/windows.rs
```diff
@@ -987,6 +1088,253 @@ pub mod tests {
...
+        ));
+        let access_denied = crate::Error::io(std::io::Error::from_raw_os_error(
+            ERROR_ACCESS_DENIED as i32,
+        ));
+
+        assert!(super::is_notify_enum_dir_error(&notify_enum_dir));
+        assert!(!super::is_notify_enum_dir_error(&access_denied));
+    }
+
+    fn encoded_record(next_entry_offset: u32, action: u32, file_name: &[u16]) -> Vec<u8> {
```
dfaust: `next_entry_offset` is hard to get right. I know it's due to the existing wire format. But still, please add a short description of how it works.

notify/src/windows.rs
```diff
@@ -987,6 +1088,253 @@ pub mod tests {
...
+        assert_eq!(
+            super::parse_records(&backing_buffer[..transferred_len]),
+            Ok(vec![super::FileNotifyRecord {
+                action: FILE_ACTION_ADDED,
+                file_name: vec!['a' as u16],
+            }])
+        );
+    }
+
+    fn injected_completion_events(
```
dfaust: This is a complicated function, it deserves some comments. What does it do? How does it work? How are the channels wired up?

dfaust (changes requested):
There are a few open questions and I would appreciate some more comments.

Also, while you're at it, please import at least the most basic types (like `Event` and `EventKind`) in the `tests` module. As it is right now, it's a wild mix of imports and full paths.

The tests only cover internals, no FS interaction. I guess that's unavoidable in this case, but I would prefer if somebody with a Windows system could have a look at it, too.

srothgan:
@dfaust Thanks for the review. Yeah, the previous attempt was a bit messy and grew far beyond what this bug needed. I started over and focused only on the Windows issue. The new fix is much smaller and should be easier to review and validate. Sorry for making the first pass more work than it needed to be.

srothgan:
@dfaust i dont want to be annoying but could you or a different maintainer give the new version a review or at least run the checks? This is mich smaller and to the point of the issue i am trying to resolve

dfaust (approved):

dfaust:
Yeah, this pass is much more focused. Thanks.

PR #967, opened 2026-07-27
feat: Expose option to use inotify `IN_DONT_FOLLOW`

### LINKED ISSUE #255: Optionally following symlinks
Hey folks! Have been happily using `5.0.0.pre2` for a few months now.

Our use of the `notify` crate is symlink aware, so before running `read_link`, we ensure that we use this crate's API to watch it. The behavior of `read_link` is to return the destination of a symlink, _even if it is dead_, and we rely on that in order to track the existence of the symlink and watch it in case it changes.

But on linux, the `notify` crate's `inotify` implementation currently doesn't set [IN_DONT_FOLLOW](https://linux.die.net/man/7/inotify):
```
    IN_DONT_FOLLOW (since Linux 2.6.15) 

Don't dereference pathname if it is a symbolic link. 
```
... which means that the attempt to watch a dead symlink fails.

----

To work around this, I have a [small patch](https://github.com/notify-rs/notify/commit/64880f0662db2b5ecbf25f1cccdca64bb8fac1bc) to the crate that currently just... unconditionally sets `IN_DONT_FOLLOW` (which is clearly not landable as is), and then another landable patch to avoid stating things unnecessarily: #256.  But I'd like to help ensure that this can be supported upstream, and would be interested in suggestions for how it could be included into the API as an optional toggle.

A global toggle per-watcher would be sufficient for our usecase, but it's also possible that some callers would prefer this on a watch-by-watch basis. Alternatively, perhaps a `Watcher::symlink_watch` method that was the spiritual equivalent of https://doc.rust-lang.org/std/path/struct.Path.html#method.symlink_metadata... ie, not reading/watching through a symlink if it existed?

tobni:
## Description

Adds `Watcher::watch_with` and `WatchPathConfig::with_dereference_symlinks`.

Defaulted trait method falling through to watch, so only inotify acts on it. Dangling symlinks become watchable as a side effect. Only works for non-recursive watches.

## Related Issues
Closes #255


tobni:
I do not believe the CI failure is related to the changes herein.

Edit: I have nevertheless likely found the reason for the flake and pushed what I believe to be hardening the retry.

JohnTitor:
Could you drop the changes around fsevent? I'd keep the PR purpose single, it should be done on another PR (I'm happy to take care of retrying failures so no worries for now).

tobni:
> Could you drop the changes around fsevent? I'd keep the PR purpose single, it should be done on another PR (I'm happy to take care of retrying failures so no worries for now).

Done!

tobni:
@JohnTitor took the liberty of pushing the test fixes required after the merge commit.

JohnTitor (approved):
Added some tweaks but I'm still not sure if this is the best shape. Let's try out this in the v9 rc and gather feedback! Thanks!

PR #970, opened 2026-08-01
fix: never abandon a recursive watch, report what failed instead

martinpitt:
A subdirectory whose watch could not be installed previously ended the whole recursive add: every directory the walk had not reached yet was left eternally unwatched. Nothing retried it and nothing reported it, and the only symptom was a program that stopped seeing changes in part of its tree.

To fix this, carry on with the rest of the walk and hand each failure to the event handler, so that callers can tell their coverage is incomplete and act on it. The event loop reports failures for the same reason, rather than dropping everything that is not the watch limit.

Watching a tree that holds one unreadable directory succeeds now, covering everything else, where it used to fail as a whole. That is the treatment a directory appearing later already got, except that its failure was dropped rather than reported.

Cover both cases with tests.

----

I agree.

Done.

I'll watch out for them. I covered both fixes with tests, and they fail without the fix.

Done, they pass.


martinpitt:
FTR, I ran into this with the [buck2](https://buck2.build/) build system. This will need a corresponding half to fix this for good (rebuilding the watch trees that failed, see https://github.com/facebook/buck2/pull/1434). This was several hours of debugging.. there was an unreadable directory appearing during some split second, which cut off an entire branch of the watched tree, and from then on the daemon kept sending old data until I killed it. That effect was very subtle, I just kept wondering why my changes to the tree didn't work and I got outdated file contents.

notify/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - FEATURE: [windows] report created file/folder kinds when they can be determined [#935]
 - CHANGE: [macOS] improve FSEvents callback performance by avoiding unnecessary allocations and repeated handler locking
 - PERF: [kqueue] avoid filesystem walks for recursive kqueue unwatch
+- FIX: never abandon a recursive watch, report what failed instead
```
JohnTitor:
```suggestion
- FIX: [inotify] never abandon a recursive watch, report what failed instead
```

notify/src/inotify.rs
```diff
@@ -670,10 +671,14 @@ impl EventLoop {
...
+                // The requested path itself failing is the caller's problem.
+                Err(err) if watch_self => return Err(err),
                 // TOCTOU: a subdirectory can disappear between walkdir listing it and us adding an
                 // inotify watch for it. This should not fail the overall recursive watch call.
-                Err(err) if !watch_self && matches!(err.kind, ErrorKind::PathNotFound) => {}
-                Err(err) => return Err(err),
+                Err(err) if matches!(err.kind, ErrorKind::PathNotFound) => {}
+                // Neither should anything else: returning would leave every directory the walk has
+                // not reached yet unwatched, with nothing to notice. Report it and carry on.
+                Err(err) => self.event_handler.handle_event(Err(err)),
```
JohnTitor: For recursive additions triggered by `add_watch(..., watch_self = false)` reaches this arm. The error is sent to the handler and `add_watches_for_paths` then returns `Ok(())`, so the existing `reached_limit` check in `handle_inotify` never sees it and cannot stop processing the remaining paths, it's a waste of processing time.

Could we special-case `MaxFilesWatch` before this arm and `return Err(err)`? That would preserve the existing stop-on-limit behavior while still continuing for recoverable, path-specific failures.

notify/src/inotify.rs
```diff
@@ -575,12 +575,13 @@ impl EventLoop {
...
         for path in add_watches {
             if let Err(add_watch_error) = self.add_watch(path, true, false) {
-                // The handler should be notified if we have reached the limit.
+                // The handler should be notified of every failure, not just the limit.
                 // Otherwise, the user might expect that a recursive watch
                 // is continuing to work correctly, but it's not.
-                if let ErrorKind::MaxFilesWatch = add_watch_error.kind {
-                    self.event_handler.handle_event(Err(add_watch_error));
+                let reached_limit = matches!(add_watch_error.kind, ErrorKind::MaxFilesWatch);
+                self.event_handler.handle_event(Err(add_watch_error));
```
JohnTitor: `add_watches` contains directories discovered through `CREATE` and `MOVED_TO` events. By the time this code processes such an event, the directory may already have been deleted or renamed; the inotify documentation explicitly calls out this race (see the Limitations and caveats section on https://man7.org/linux/man-pages/man7/inotify.7.html).

We should continue skipping `PathNotFound` here.

notify/src/inotify.rs
JohnTitor, line 1: Note that there is another source of partial recursive watches that your PR doesn't currently cover: errors produced by `WalkDir`. So some fail paths will be ignored continuously. You might also want to touch that part if your problem would be affected.

martinpitt:
Pushed a clean rebase against current main.

JohnTitor (changes requested):

notify/CHANGELOG.md
martinpitt: Done, and also added the PR ref. I also added a second entry for the walk errors you suggested. LMK if I should rather merge them.

Unrelated: `[#964]` a few lines up has no link definition.

notify/src/inotify.rs
martinpitt: Ugh yes, thanks for spotting! This was even worse than wasting time: once the limit is hit, every remaining directory would have handed the handler its own copy of the same error. :see_no_evil:

To make this more robust, I split out that policy decision into a separate function and `WalkFailure` type, to avoid this oversight. It's a little more intrusive that way, but I think/hope worth it. Let me know if you don't like it.

notify/src/inotify.rs
martinpitt: Agreed, fixed. The event loop consults the same policy as the recursive walk now, and `PathNotFound` maps to skipping it silently.

Note, since it is not obvious from the diff: `add_watch` stats the path before anything else, and `Error::io_watch` already maps `ENOENT` to `ErrorKind::PathNotFound`. So the created-then-deleted directory really does arrive as that kind and really is filtered.

notify/src/inotify.rs
martinpitt: You are right, and it is the same bug one layer up: `filter_dir` threw away every `walkdir::Error`, so a directory that could not be listed silently pruned that whole subtree. Fixed in a separate commit.

This requires a little dance: An unreadable directory fails both `inotify_add_watch` and `read_dir`, so for tasteful/useful error reporting these duplicates need to be filtered out. That commit does it and unit-tests it.

martinpitt:
This became conflicty, resolved/rebased.

martinpitt:
Pushed a clean rebase against current origin/main, no actual changes. I just want my branch to keep up with upstream changes.

JohnTitor (approved):
Sorry for the delay! Overall LGTM with a fix not to show unrelated (i.e. failures from not watched paths) errors during walking.

martinpitt:
Excellent, thanks @JohnTitor !

PR #971, opened 2026-08-01
fsevent: refuse to create a stream that would corrupt the process

daandemeyer:
FSEvents closes a file descriptor it does not own once a stream carries more
than roughly RLIMIT_NOFILE/10 paths. The descriptor is fd 0, which then becomes
the lowest free descriptor, so the next open anywhere in the process lands on it
and the following oversized stream closes that instead. The damage surfaces far
from FSEvents, as EBADF on unrelated files, which is why it is so hard to trace
back: it was originally found as intermittent EBADF from closedir inside a
remove_dir_all that had nothing to do with watching.

Reproduced on macOS 14, 15 and 26 with a standalone C program that uses nothing
but CoreServices, so this is not something notify is doing wrong. It is
deterministic rather than racy, and the threshold scales with the descriptor
limit: with the common default of 256 it is around 27 paths. The reproducer is
at https://github.com/daandemeyer/fsevents-fd-repro.

Compute the budget from getrlimit and return MaxFilesWatch rather than making
the call. Refusing to watch is worse for the caller than watching would have
been, but far better than silently corrupting the descriptor table of the host
process, and unlike a corrupted descriptor table it is something they can
respond to. The error message points at the three things that actually help:
raising RLIMIT_NOFILE, watching fewer paths, or the kqueue backend.

The budget uses a twelfth rather than a tenth. The exact edge shifts with how
many descriptors the process already holds, so a little headroom avoids sitting
on it.

This subsumes error_properly_on_stream_start_failure, which watched 4097 paths
to provoke an FSEventStreamStart failure. Those 4097 paths are what closes fd 0,
so the test was corrupting the process it ran in: it carried a catch_unwind
around its own temporary directory cleanup to survive the resulting EBADF, and
on CI it sometimes aborted the whole test binary instead, which catch_unwind
cannot intercept. The guard now rejects that call before FSEvents sees it, so
replace the test with one that derives an over-budget path count from the
current rlimit and asserts MaxFilesWatch. No stream is created, so the cleanup
workaround goes away with it. There is no longer a way to reach a genuine start
failure from a test without deliberately re-entering the range that corrupts the
process, so that path is left uncovered on purpose.

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -12,7 +12,7 @@
 - FIX: [windows] emit a rescan event when `ReadDirectoryChangesW` discards change details [#964]
 - FEATURE: [windows] report created file/folder kinds when they can be determined [#935]
 - CHANGE: [macOS] improve FSEvents callback performance by avoiding unnecessary allocations and repeated handler locking
-- FIX: [macOS] refuse to create an FSEvents stream carrying more paths than macOS handles without closing a file descriptor this process owns
+- FIX: [macOS] refuse to create FSEvents streams whose combined path count would make macOS close a file descriptor this process owns
 - PERF: [kqueue] avoid filesystem walks for recursive kqueue unwatch
 
 [#930]: https://github.com/notify-rs/notify/pull/930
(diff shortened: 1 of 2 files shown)


daandemeyer:
The logs one of the runs on the github repository that show the issue reproducing on the GA macos runners: https://github.com/daandemeyer/fsevents-fd-repro/actions/runs/30170828666/job/89711549736#step:4:19

daandemeyer:
I have reported it to Apple: https://feedbackassistant.apple.com/feedback/23982709 but it will likely take a lot of time to get fixed and deployed to all Apple hardware out there so I think in the meantime we should make sure we don't trigger the bug in notify.

JohnTitor (approved):
Thanks, overall LGTM. Sadly Actions is currently down so I'll get back later: https://www.githubstatus.com/incidents/qcvjkzcs7j74

PR #975, opened 2026-08-07
fsevent: coalesce watches nested inside other watches

daandemeyer:
An FSEvents stream root reports the entire hierarchy below it, and
`callback_impl` attributes every event to the longest watch it falls
under, so passing both a directory and something below it to
`FSEventStreamCreate` adds nothing while costing a slot against the path
limit that macOS enforces by corrupting the process descriptor table
above roughly RLIMIT_NOFILE/10 paths.

Derive the CFArray handed to `FSEventStreamCreate` from the watch map in
`run()` instead of maintaining it alongside, dropping any watch whose
ancestors include another watch on the same volume. FSEvents does not
promise that a stream root reports events from a volume mounted below
it, so a covering ancestor across a mount point does not count.
Unwatching a covering directory promotes the nested watches back to
stream roots on the next run with no separate bookkeeping, and
`remove_cf_path` goes away.

One behavioural difference: with `kFSEventStreamCreateFlagWatchRoot`, a
nested watch no longer receives a root-changed event of its own when an
ancestor is renamed. A recursive covering watch still reports the
rename; a non-recursive one reports nothing, since `callback_impl`
accepts only direct children for those.

Covered by unit tests for the stream roots produced by nested and by
sibling watches, and by black-box tests that a nested watch still
delivers events after its covering watch is removed and that a covering
watch still delivers events outside the nested one.

### DIFF
--- a/notify/CHANGELOG.md
+++ b/notify/CHANGELOG.md
@@ -13,7 +13,7 @@
 - FEATURE: [windows] report created file/folder kinds when they can be determined [#935]
 - CHANGE: [macOS] improve FSEvents callback performance by avoiding unnecessary allocations and repeated handler locking
 - FIX: [macOS] refuse to create FSEvents streams whose combined path count would make macOS close a file descriptor this process owns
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Makes sense 👍 added a tweak around some coalescing condition.

PR #979, opened 2026-08-11
poll: stop the tests from depending on second boundaries

daandemeyer:
PathData::mtime has whole-second resolution and is compared before the content hash, so the kind of event a change produces depends on whether it crossed a wall-clock second. Six poll tests asserted one of the two outcomes and failed about one run in ten under the load of the full suite.

Mutating a watched directory bumps its own write time, so create_file, create_dir, remove_file and rename_file can get an extra Modify(Metadata(WriteTime)) for the directory; accept it as optional. modify_file and create_write_overwrite leave an entry with both new contents and a new write time, reported as Data within a second and as Metadata(WriteTime) across one. Set that entry's write time an hour ahead before the baseline scan and the comparison always falls through to the hash. Restoring the old write time after the change is not enough: on macOS the scan can still stat the write time of the change.

poll-watcher-hashing.rs did restore it, but polled every 10ms, so a scan could land between the write and the restore. Poll it manually instead and skip the directory's own event. Trailing optional expectations also needed a harness fix, since is_empty counted a queue holding only optionals as non-empty and the waiter blocked until it timed out.

Sleeping a second before each mutation fails all six tests before this change and passes after.


JohnTitor (approved):
Makes sense, thanks!

PR #980, opened 2026-08-13
refactor(file-id): reduce unsafe block range and add safety comment

shulaoda:
## Description

`get_file_info` and `get_file_info_ex` are declared as `unsafe fn`, but they have no preconditions for callers, so every call site ends up wrapping them in an `unsafe` block for no reason.

This PR makes them safe functions and keeps `unsafe` only around the actual FFI calls, each with a `SAFETY` comment. It also uses `MaybeUninit` instead of `mem::zeroed()` and removes the unneeded `as HANDLE` casts.

Ported from the rolldown fork: rolldown/notify#52 (by @sapphi-red).

## Additional review notes

Rolldown currently maintains a fork of notify ([rolldown/notify](https://github.com/rolldown/notify)). We'd like to start upstreaming its small, reasonable changes like this one, and propose some more general features later, so that rolldown can eventually switch back to upstream notify instead of maintaining the fork long-term.

### DIFF
--- a/file-id/src/lib.rs
+++ b/file-id/src/lib.rs
@@ -150,7 +150,7 @@ pub fn get_file_id_no_follow(path: impl AsRef<Path>) -> io::Result<FileId> {
 pub fn get_file_id(path: impl AsRef<Path>) -> io::Result<FileId> {
     let file = open_file(path)?;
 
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Looks good, thanks!

PR #981, opened 2026-08-13
PollWatcher: detect subsecond mtime changes

marcoshernanz:
## Description

PollWatcher stores file modification times at whole-second precision. When content comparison is disabled, multiple writes within the same second compare equal and the later write produces no event.
This keeps nanosecond precision for ordinary files so those writes are detected. Directories and content-hashed watches retain their previous whole-second behavior to avoid redundant directory events and preserve existing event classification.
A regression test advances a file mtime twice within one second and verifies a WriteTime event is emitted without content hashing.

## Related Issues
- https://github.com/vercel/next.js/issues/96982

### DIFF
--- a/notify/src/poll.rs
+++ b/notify/src/poll.rs
@@ -90,10 +90,12 @@ mod data {
 
     use super::ScanEventHandler;
 
-    fn system_time_to_seconds(time: std::time::SystemTime) -> i64 {
+    const NANOS_PER_SECOND: i128 = 1_000_000_000;
+
+    fn system_time_to_nanos(time: std::time::SystemTime) -> i128 {
         match time.duration_since(std::time::SystemTime::UNIX_EPOCH) {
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Makes sense, thanks!

notify/src/poll.rs
```diff
@@ -402,10 +404,19 @@ mod data {
         /// Create a new `PathData`.
         fn new(data_builder: &DataBuilder, meta_path: &MetaPath) -> PathData {
             let metadata = meta_path.metadata();
+            let is_file = metadata.is_file();
+            let mut mtime = metadata.modified().map_or(0, system_time_to_nanos);
+            if !is_file || data_builder.compare_contents {
```
sokra, line 409: This condition also includes symlinks as they are `!is_file`. So symlink precision is still "broken" as in having 1s precision

PR #982, opened 2026-08-15
inotify,kqueue: extract test_event_loop helpers

daandemeyer:
Both backends' tests built their event loop inline, five times each. The
repetition is noise in every test that follows, and it means a later
change to how a test event loop is constructed has to be made ten times.

Extract one helper per backend. No behavior change.


JohnTitor (approved):
Thanks!

PR #987, opened 2026-08-17
perf(windows): reduce fs::metadata calls when adding a watch

shulaoda:
## Description

On Windows, `watch_inner` checks that the path is a file or a directory before sending the request to the watcher thread, and `add_watch` then performs the exact same check again, plus one more `is_dir` call to decide between file and directory watching. Each `is_dir`/`is_file` call is a separate `fs::metadata` syscall, so a single watch request ends up doing up to five of them.

This PR removes the duplicated client-side check and reuses one `metadata` result in `add_watch`, reducing it to a single call. This mainly matters for applications that watch many individual files. As a side effect, the error for a non-existent path now includes the path, since it is reported by the watcher thread.

Ported from the rolldown fork: rolldown/notify@d44e3277f166ee58b07aee3cefd204bedd30015d.

## Additional review notes

Part of the same effort as #980: upstreaming small changes from the rolldown fork.

## Related Issues

None.

### DIFF
--- a/notify/src/windows.rs
+++ b/notify/src/windows.rs
@@ -287,8 +287,13 @@ impl ReadDirectoryChangesServer {
         is_recursive: bool,
         separator_style: SeparatorStyle,
     ) -> Result<PathBuf> {
+        let metadata = path.absolute.metadata();
         // path must exist and be either a file or directory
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Cool, thanks!

PR #989, opened 2026-08-20
feat(freebsd): select inotify on 14.5+

kev009:
## Description

Follow-up to #972: native inotify ships in FreeBSD 14.5 (syscalls and libc wrappers) thanks to @markjdb, not only 15.0+.

Autoselect now compares `freebsd-version -u` as `(major, minor) >= (14, 5)` so native builds on 14.5+ use inotify and older 14.x stays on kqueue. The `freebsd_inotify` feature still forces inotify for cross-compiles targeting 14.5+.

CI is unchanged: 14.4 covers kqueue, 15.1 covers inotify as it will be some time before the 14.4 builder ages out and the coverage testing is still useful.

## Related Issues

- #972
- #952

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -63,7 +63,7 @@ body:
       description: Which backend is being used?
       options:
         - Unknown/Auto-detected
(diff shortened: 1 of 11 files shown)


JohnTitor (approved):
Thanks!

